PR #1, opened 2025-08-01
copy edge-impulse lib from nrf

Szynkaa:
All files under `lib/edge_impulse` and `tests/lib/edge_impulse` are just copy from nrf. Other files contains needed structure in Kconfig/CMakeLists and patch for nrf to hide its edge-impulse lib.


Szynkaa:
invalid

PR #5, opened 2025-08-08
applications: machine_learning: add Neuton and profiling in ML app

Szynkaa:
Extract EI Wrapper part from ml_runner and create separate runners for: Edge Impulse (via wrapper), Neuton and stub for baseline. Documentation update will follow in separate commit.


applications/machine_learning/src/runners/Kconfig.stub_runner
```diff
@@ -0,0 +1,36 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menuconfig ML_APP_ML_RUNNER_STUB
+	bool "Neuton based machine learning model"
+	select FPU
+	depends on CAF_SENSOR_EVENTS
+	select ML_APP_ML_RESULT_EVENTS
+	imply CBPRINTF_FP_SUPPORT
```
dchat-nordic: This only implies the libc nano support. I would use the "REQUIRES_FLOAT_PRINTF" instead, but please look into it.

applications/machine_learning/src/runners/neuton_runner.c
```diff
@@ -0,0 +1,114 @@
...
+
+int runner_start_prediction(void)
+{
+	return 0;
+}
+
+int runner_stop_prediction(void)
+{
+	return 0;
+}
```
dchat-nordic: I would consider using ENOSYS here since the functions do nothing. However, keep it as a suggestion if this would require a lot of changes.

applications/machine_learning/src/runners/runner.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include <stddef.h>
+#include <stdbool.h>
+
+typedef bool (*runner_continue_callback)(int err);
+
+int runner_init(const runner_continue_callback cb);
+int runner_start_prediction(void);
+int runner_stop_prediction(void);
+int runner_add_data(const float *data, size_t data_size);
```
dchat-nordic: I know this is the internal API, but since your PR adds the possiblity of using multiple runners, it would be beneficial to define a simple API for them. Please add a short note about the functions and their input and return values. This doesn't need to be very long, just a brief is ok.

dchat-nordic (changes requested):
I am not sure about the copyright notices. AFAIK new files get 2025, but old (changed) files should not get a range ie. 2023-2025, instead their original value. I might be wrong, let's check this with someone. Maybe @peknis will know this.
Otherwise the PR looks good.

applications/machine_learning/src/modules/inference_position.c
```diff
@@ -0,0 +1,53 @@
...
+#include <caf/events/sensor_event.h>
+#include <hal/nrf_gpio.h>
+
+#define ZEPHYR_USER_NODE DT_PATH(zephyr_user)
+
+#define ABS_GPIO_PIN_BY_IDX(node_id, prop, idx)                                                    \
+	NRF_GPIO_PIN_MAP(DT_PROP(DT_PHANDLE_BY_IDX(node_id, prop, idx), port),                     \
+			 DT_GPIO_PIN_BY_IDX(node_id, prop, idx))
+
+static const uint32_t pins[] = {DT_FOREACH_PROP_ELEM_SEP(
```
grochu, line 24: Where are the pins defined in DTS? Can this module be sure there will be at least 2 gpios provided?

applications/machine_learning/src/runners/neuton_runner.c
Szynkaa: They cannot return ENOSYS as this means error for caller at applications/machine_learning/src/modules/ml_runner.c:56

```
		err = runner_start_prediction();
		if (err) {
			report_error();
		}
```

dchat-nordic: Please look at https://github.com/nrfconnect/sdk-nrf/blob/main/applications/ipc_radio/src/main.c which handles just the case with ENOSYS.

applications/machine_learning/src/runners/runner.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifndef _RUNNER_H_
+#define _RUNNER_H_
+
+#include <stddef.h>
+#include <stdbool.h>
+
+typedef bool (*runner_continue_callback)(int err);
+
+int runner_init(const runner_continue_callback cb);
```
grochu, line 30: Please add function descriptions (dox -style).

applications/machine_learning/src/runners/neuton_runner.c
```diff
@@ -0,0 +1,114 @@
...
+			continue_cb(err);
+		}
+
+		neuton_u16_t predicted_target;
+		const neuton_output_t *probabilities;
+
+		const neuton_i16_t targets_num =
+			neuton_nn_run_inference(inference_input, &predicted_target, &probabilities);
+
+		const int err = (targets_num < 0) ? targets_num : 0;
```
grochu, line 60: Why not define a single var `err` (another one in the `if` block, line 49)?

grochu (commented):
I hardly know this application, so hard to do a full review. A few question to begin with:

applications/machine_learning/src/modules/inference_position.c
Szynkaa: Pins will be defined in in app.overlay, when configuration using this module will be added.
Bound checking done.

applications/machine_learning/src/runners/Kconfig.stub_runner
Szynkaa: Removed entirely

applications/machine_learning/src/runners/neuton_runner.c
Szynkaa: done

applications/machine_learning/src/runners/runner.h
Szynkaa: done

applications/machine_learning/src/runners/runner.h
Szynkaa: done

applications/machine_learning/src/runners/neuton_runner.c
Szynkaa: They are not connected. This will add extra line and not reduce cognitive complexity

applications/machine_learning/src/runners/stub_runner.c
```diff
@@ -0,0 +1,76 @@
...
+
+	count += data_size / FRAME_SIZE;
+
+	if (count >= WINDOW_SIZE) {
+		count -= WINDOW_SIZE;
+		k_sem_give(&inference_sem);
+	}
+
+	return 0;
+}
```
dchat-nordic, line 76: nit: There is no check to guard against data_size being divisible by frame size.

dchat-nordic (approved):

grochu (approved):

PR #10, opened 2025-11-21
Release/nrf edge ai lab launch

rrusak:
- Removed Edge Impulse integration (`ei_wrapper`, samples, tests)
- Added Nordic Edge AI Lab C runtime library
- Added 3 reference samples: regression, classification, anomaly detection
- Added comprehensive documentation and library API docs
- All samples build with `west` and work on supported Nordic boards


doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,118 @@
+.. _ug_nrf_edgeai:
+
+NRF Edge AI integration
+#######################
+
+.. contents::
+   :local:
+   :depth: 2
+
+The NRF Edge AI add-on provides a lightweight wrapper that enables deploying
```
wbober: ```nRF``` all around

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,118 @@
...
+machine learning models on Nordic devices using the Nordic Edge AI Lab runtime.
```
wbober: Use links to the edge ai lab docs.

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,118 @@
...
+
+Typical integration steps:
+
+1. Prepare or generate a model using the Nordic Edge AI Lab web tooling.
+2. Place the generated source files into the application (for example
+   ``samples/nrf_edgeai/<sample>/src/nrf_edgeai_generated``).
+3. Replace samples user data, feeding and result handling code with application-specific
+   logic as required by your generated solution.
+4. Enable the integration in the application configuration (see the
+   Kconfig options) and, if required, select a full libc implementation.
```
wbober: How do I know if full libc is required?

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,118 @@
...
+.. include:: ../includes/classification_integration_examples_edgeai.txt
+
+Regression:
+
+.. include:: ../includes/regression_integration_examples_edgeai.txt
+
+Anomaly detection:
+In anomaly detection mode, the model's inference yields an Anomaly Score, 
+indicating the similarity of input data to the "normal" data used for training.
+A higher Anomaly Score signifies greater deviation from the normal data,
```
wbober: Is it a value normalized between 0-1?

wbober (commented):
First pass. 

- [x] Could you please squash all changes related to removing existing state into a single commit, so we can revert later on?
- [x] I haven't spotted any mention of the DSP lib. I think it would be worth to do that?

rrusak:
> First pass.
> 
> * [ ]  Could you please squash all changes related to removing existing state into a single commit, so we can revert later on?
> * [ ]  I haven't spotted any mention of the DSP lib. I think it would be worth to do that?

Unfortunately, in some commits in which I added new files, I've also removed some of those related to edge impulse. I can squash the majority, but this will not allow one commit to revert all the changes concerning the removal of edge impulse. 
Or I can roll back this PR to the current master, then remove everything related to edge impulse, and then add the changes to nrf_edgeai again. But I think that if we want to return edge impulse to the addon, the addon will look different, and these changes will still have to be edited after the revert.  We always have history, and we can manually pull these changes and roll them back.

Anyway, the choice is yours :) what can I do: 
- squash most of it, but not all of it. 
- roll back the changes and re-do it with a single (or squashed) commit for edge impulse removing. 
- or leave it as is.


Regarding DSP... do we want to position it now as a separate library for use separately from the edgeai runtime library? Personally, I wouldn't want to focus too much on this right now, as there are a number of issues in some functions that I'd like to address before promoting it to a wider audience....  There's doxygen API documentation in the library itself, but I could add some more general information in the rst files.

doc/integrations/nrf_edgeai.rst
rrusak: The linker will return an error when trying to link expf or sqrtf... If the user model is floating point and not quantized, or some DSP functions require sqrtf, we need to link libm (<math.h>)

doc/integrations/nrf_edgeai.rst
rrusak: No, it's not normalized because when we train on normal data, we don't know all the degrees of anomalous data and therefore can't define normalization coefficients. Our anomaly score is a mathematical distance calculated using the cosine similarity formula; it can be higher than 1.

rrusak: added explanations on how to interpret the Anomaly Score and validation strategy, and how to set the anomaly threshold.

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,165 @@
+.. _ug_nrf_edgeai:
+
+nRF Edge AI integration
```
wbober: I think we need to settle on some semantics.

One direction is to name this page "Nordic Edge AI Lab" integration. And then talk how "nRF Edge AI Runtime" enables running models produced by Nordic Edge AI Lab.

Another direction is to simply talk about "nRF Edge AI Runtime".

For me a "runtime" is always the embedded code.

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,165 @@
...
+#######################
+
+.. contents::
+   :local:
+   :depth: 2
+
+The nRF Edge AI add-on provides a lightweight wrapper that enables deploying
```
wbober: `Add-On`

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,165 @@
...
+machine learning models on Nordic devices using the Nordic Edge AI Lab runtime. 
+In this documentation the underlying platform is referred to as `Nordic Edge AI Lab`_.
+
+Integration prerequisites
+*************************
+
+Before you start using the nRF Edge AI integration, ensure the following are in place:
```
wbober: Nordic Edge AI Lab?

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,165 @@
...
+
+* Installation of the nRF Connect SDK and the corresponding Zephyr toolchain.
```
wbober: Simply refer to NCS installation instructions from the NCS docs.

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,165 @@
...
+   - ``thingy53/nrf5340/cpuapp``
+
+   See the individual samples' ``sample.yaml`` files under
+   ``samples/nrf_edgeai/<sample>/sample.yaml`` for exact supported platforms
+   and CI test coverage.
+
+Solution architecture
+*********************
+
+The integration is implemented as a wrapper library (`lib/nrf_edgeai`) that
```
wbober: I would reduce the number of terms used. Runtime is a library.

doc/integrations/nrf_edgeai.rst
```diff
@@ -0,0 +1,165 @@
...
+adapts models and runtime helpers from the `Nordic Edge AI Lab`_ to the nRF
+Connect SDK build system. Models are typically distributed as generated C
+sources (header + C files) which are consumed by the sample applications.
+
+The wrapper exposes a small API that abstracts model initialization, input
```
wbober: ```The runtime``` exposes..."?

doc/libraries/nrf_edgeai.rst
```diff
@@ -0,0 +1,197 @@
+.. _nrf_edgeai_lib:
+
+nRF Edge AI Library
```
wbober, line 3: nRF Edge AI Runtime?

doc/libraries/nrf_edgeai.rst
```diff
@@ -0,0 +1,197 @@
...
+   :depth: 2
+
+The nRF Edge AI library provides a lightweight C-based runtime for deploying machine learning models on Nordic Semiconductor devices.
+Models are generated using the `Nordic Edge AI Lab`_ web tooling and distributed as C source files that integrate seamlessly with the nRF Connect SDK build system.
+For more information about integrating nRF Edge AI models, see :ref:`ug_nrf_edgeai`.
+
+Overview
+********
+
+The library consists of several core modules:
```
wbober, line 15:
```
The runtime consists of the following modules:
**Core** - inference engine...
```

?

wbober (changes requested):
Individual *.rst pages for samples seem to be missing. See https://docs.nordicsemi.com/bundle/addon-sidewalk-latest/page/samples/variants/hello.html#variant-sidewalk-hello as a reference.

doc/integrations/nrf_edgeai.rst
rrusak: Sure
- Nordic Edge AI Lab - web platform
- nRF Edge AI Runtime - embedded code, part of nRF Edge AI library to run generated models from Nordic Edge AI Lab.
- Nordic Edge AI Lab generate nRF Edge AI runtime solution (model?) 
- Embedded developer integrates and using nRF Edge AI library or/and  nRF Edge AI runtime library

What do you think? For me runtime is part of nRF Edge AI library (because there is separated DSP/NN/<some_module> block), for me runtime is a part of library to run Nordic Edge AI Lab model.

Example:
The TensorFlow C/C++ “runtime” refers to the core components needed to run TensorFlow models (graphs, ops, kernels, memory allocators, execution engine), without the front-end.

It is not the whole TensorFlow project — it is the low-level execution engine + a set of compiled ops.

So "runtime" term is not very clear, but let's settle on something. Just take into account that in the C code, I've separeted nrf_edgeai/rt/ folder for "runtime"...

doc/integrations/nrf_edgeai.rst
rrusak: nRF Edge AI library? CMSIS also uses term "compute library"

rrusak:
> Individual *.rst pages for samples seem to be missing. See https://docs.nordicsemi.com/bundle/addon-sidewalk-latest/page/samples/variants/hello.html#variant-sidewalk-hello as a reference.

done

include/nrf_edgeai/dsp/fast_math/nrf_dsp_cmplx_math.h
```diff
@@ -0,0 +1,53 @@
...
+#include <nrf_edgeai/dsp/nrf_dsp_types.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Computes the magnitude of the elements of a complex data vector.
+ * 
+ * @param[in]   p_input          Input array and the data is stored in an interleaved fashion [xR0, xI0, xR1, xI1, ..., xRn, xIn].
```
Szynkaa, line 26: Please try to stick to Zephyr coding rules used in NCS. Here for example you exceed 100 characters.

include/nrf_edgeai/dsp/platform/nrf_dsp_platform.h
Szynkaa, line 1: minor: missing empty line at the end of some files

include/nrf_edgeai/dsp/nrf_dsp.h
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2025 Nordic Semiconductor ASA
+ * SPDX-License-Identifier: Apache-2.0
+ */
+/**
+ *
+ * @defgroup nrf_dsp Nordic Semiconductor Digital Signal Processing SDK
+ * @{
+ * @ingroup nrf_dsp
+ *
```
Szynkaa, line 10: minor: many empty lines and end of docs comments

include/nrf_edgeai/dsp/nrf_dsp_types.h
```diff
@@ -0,0 +1,222 @@
...
+} nrf_dsp_stat_ctx_i16_t;
+
+/**
+ * @brief INT16-fixed-point spectral reusable context, in which is used to store derived or intermediate
+ *        metrics to simplify / speed up the calculation of dependent metrics
+ * 
+ */
+typedef struct nrf_dsp_spectral_ctx_i16_s
+{
+
```
Szynkaa, line 168: nitpick: unneeded empty line

include/nrf_edgeai/dsp/platform/nrf_dsp_platform_types.h
```diff
@@ -0,0 +1,20 @@
...
+#include <nrf_edgeai/nrf_edgeai_ctypes.h>
+#else
+// User-defined platform types
+#include <nrf_dsp_custom_platform_types.h>
+#endif  // NRF_DSP_CUSTOM_PLATFORM_TYPES
+
+// Header for defining platform-specific types for DSP library:
+// typedef float flt32_t;
+// typedef double flt64_t;
+// typedef size_t size32_t;
```
Szynkaa, line 18: Should this be here?

include/nrf_edgeai/dsp/private/nrf_dsp_c_intrinsics.h
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * @brief   Rotate Right in unsigned value (32 bit)
+ * @details Rotate Right (immediate) provides the value of the contents of a register rotated by a variable number of bits.
+ * @param [in]    op1  Value to rotate
+ * @param [in]    op2  Number of Bits to rotate
+ * @return               Rotated value
+ */
+__NRF_DSP_STATIC_FORCEINLINE uint32_t __ROR(uint32_t op1, uint32_t op2)
```
Szynkaa, line 93: Give op1 and op2 meaningful names

include/nrf_edgeai/dsp/private/nrf_dsp_c_intrinsics.h
```diff
@@ -0,0 +1,138 @@
...
+    return ((int32_t)(x >> 32) != ((int32_t)x >> 31)) ? ((0x7FFFFFFF ^ ((int32_t)(x >> 63)))) :
+                                                        (int32_t)x;
+}
+
+/**
+ * @brief Clips INT64 to INT16 values.
+ */
+__NRF_DSP_STATIC_FORCEINLINE int16_t clip_i64_to_i16(int64_t x)
+{
+    return ((int32_t)(x >> 32) != ((int32_t)x >> 31)) ? ((0x7FFF ^ ((int16_t)(x >> 63)))) :
```
Szynkaa, line 113: I suspect that, there should be shift by 16 and 15 bits in ternary's condition

include/nrf_edgeai/dsp/fast_math/nrf_dsp_cmplx_math.h
rrusak: Hi! at the moment nrf_edgeai library does not support NCS/Zephyr coding rules, there was no requirement for that, maybe in the future it will, but not in this version I belive :)

include/nrf_edgeai/nn/platform/nrf_nn_platform_types.h
```diff
@@ -0,0 +1,20 @@
...
+#include <nrf_edgeai/nrf_edgeai_ctypes.h>
+#else
+// User-defined platform types
+#include <nrf_nn_custom_platform_types.h>
+#endif  // NRF_NN_CUSTOM_PLATFORM_TYPES
+
+// Header for defining platform-specific types for NN library:
+// typedef float flt32_t;
+// typedef double flt64_t;
+// typedef size_t size32_t;
```
Szynkaa, line 18: Needed?

include/nrf_edgeai/nrf_edgeai_ctypes.h
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * @brief 64-bit floating-point type definition.
+ */
+typedef double flt64_t;
+
+/**
+ * @brief Architecture 32-bit size_t type definition.
+ */
+typedef size_t size32_t;
```
Szynkaa, line 30: How do we assure that size_t is actually 32-bit?

samples/nrf_edgeai/anomaly/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+# Zephyr CMake project
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(nrf_edgeai_classification_sample)
+
+# Source files
+target_include_directories(app PRIVATE include)
+target_sources(app PRIVATE
+src/main.c
+src/nrf_edgeai_generated/nrf_edgeai_user_model.c
```
Szynkaa: indent

samples/nrf_edgeai/anomaly/prj.conf
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
```
Szynkaa:
```suggestion
# Copyright (c) 2025 Nordic Semiconductor ASA
```

samples/nrf_edgeai/anomaly/sample.yaml
```diff
@@ -1,8 +1,8 @@
 sample:
-  description: Sample showing Edge Impulse data forwarder usage
-  name: Edge Impulse data forwarder sample
+  description: Sample showing NRF Edge AI Anomaly Detection model usage and results interpretation.
+  name: NRF Edge AI Anomaly Detection Sample
```
Szynkaa:
```suggestion
  description: Sample showing nRF Edge AI Anomaly Detection model usage and results interpretation.
  name: nRF Edge AI Anomaly Detection Sample
```

samples/nrf_edgeai/anomaly/src/main.c
Szynkaa, line 1: License header missing

samples/nrf_edgeai/classification/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+# Zephyr CMake project
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(nrf_edgeai_classification_sample)
+
+# Source files
+target_include_directories(app PRIVATE include)
+target_sources(app PRIVATE
+src/main.c
+src/nrf_edgeai_generated/nrf_edgeai_user_model.c
```
Szynkaa: indent

samples/nrf_edgeai/anomaly/Kconfig.sysbuild
Szynkaa, line 1: is this Kconfig necessary across samples?

samples/nrf_edgeai/regression/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+# Zephyr CMake project
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(nrf_edgeai_regression_sample)
+
+# Source files
+target_include_directories(app PRIVATE include)
+target_sources(app PRIVATE
+src/main.c
+src/nrf_edgeai_generated/nrf_edgeai_user_model.c
```
Szynkaa: indent

samples/nrf_edgeai/classification/src/main.c
```diff
@@ -0,0 +1,465 @@
...
+{
+	nrf_edgeai_err_t res;
+
+	/* Feed samples point-by-point to match real-world streaming scenario */
+	for (size_t i = 0; i < data_len; i++) {
+		/* Accumulate single sample into the model's input window */
+		flt32_t input_sample = p_input_data[i];
+		res = nrf_edgeai_feed_inputs(p_user_model, &input_sample, 1 * USER_UNIQ_INPUTS_NUM);
+
+		if (res == NRF_EDGEAI_ERR_SUCCESS) {
```
Szynkaa, line 326: maybe remove indent by inverting logic? applies also to `model_predict` in other samples

Szynkaa (commented):
I mostly ignored function declarations for dsp

 In all new samples:
- main.c is missing license header
- samples.yaml us NRF instead of nRF
- prj.conf use wrong year in header

include/nrf_edgeai/dsp/private/nrf_dsp_c_intrinsics.h
rrusak: No the shift should be by 15 bits at the end, because the function is doing a saturating rounding conversion, not a plain right-shift. The function does : "Take a 64-bit fixed-point number in Q15 format and convert it to a 16-bit signed integer with saturation".

In fixed-point terminology:

x is treated as a Q(16.15) or similar format.

The function extracts the top 16 bits of (x >> 15).

If you instead shifted by 16 bits, you'd simply drop 16 bits and effectively extract the top 16 bits of the 64-bit number — but that would be asymmetric, and would shift the sign boundary incorrectly.

To fit cleanly into int16_t (−32768…32767), the allowed range for x must be:

INT16_MIN << 15 = -32768 * 32768 = -1073741824
INT16_MAX << 15 =  32767 * 32768 =  1073709056

These numbers match the threshold in the clipping logic — confirming that the function is designed around a (value × 2¹⁵) scaling.

So:
- x >> 15 extracts the unscaled integer
- saturation occurs if that integer would overflow 16 bits

samples/nrf_edgeai/classification/src/main.c
rrusak: I dont know if it will be better.... I would like to have clear logic for users that you need to wait success NRF_EDGEAI_ERR_SUCCESS to continue, if you do invert logic (res != NRF_EDGEAI_ERR_SUCCESS) then you should handle different not success states :) and I don't want to overload with handling errors or other states

samples/nrf_edgeai/anomaly/src/main.c
```diff
@@ -0,0 +1,350 @@
...
+		/* Extract one sample pair: [X_acceleration, Y_acceleration] */
+		flt32_t input_sample[USER_UNIQ_INPUTS_NUM];
+		input_sample[0] = p_input_data[i];     /* Sensor 1: X-axis */
+		input_sample[1] = p_input_data[i + 1]; /* Sensor 2: Y-axis */
+
+		/* Feed this sample pair into the model's windowing buffer */
+		res = nrf_edgeai_feed_inputs(p_user_model, input_sample, 1 * USER_UNIQ_INPUTS_NUM);
+
+		if (res == NRF_EDGEAI_ERR_SUCCESS) {
+			/* Input buffer has reached 128 samples—run inference on the complete window
```
mbarchie: An em dash character may cause issues sometimes, can be replaced with a regular hyphen-minus
```suggestion
			/* Input buffer has reached 128 samples - run inference on the complete window
```

include/nrf_edgeai/dsp/private/nrf_dsp_c_intrinsics.h
Szynkaa: Your response is hard to follow for me as I have low experience with Q notation and similar things.

But I run some tests cases and I think it should be `((int32_t)(x >> 16) != ((int32_t)x >> 15)) ? ((0x7FFF ^ ((int16_t)(x >> 63)))) : (int16_t)x`

samples/nrf_edgeai/anomaly/src/main.c
```diff
@@ -0,0 +1,350 @@
...
+ *   5. Compute anomaly score based on learned feature representation
+ *   6. Return single score indicating anomaly likelihood
+ *
+ * Return:
+ *   Anomaly score (float): 0.0 = definitely healthy
+ *   Interpretation:
+ *     score < USER_ANOMALY_THRESHOLD -> Healthy gear (pass maintenance check)
+ *     score >= USER_ANOMALY_THRESHOLD > Anomalous gear (schedule maintenance alert)
+ */
+flt32_t model_predict(nrf_edgeai_t *p_user_model, const flt32_t *p_input_data, size_t data_len)
```
mbarchie: Function can be made "static" to limit its scope

samples/nrf_edgeai/anomaly/src/main.c
```diff
@@ -0,0 +1,350 @@
...
+{
+	nrf_edgeai_err_t res;
+
+	flt32_t anomaly_score = 10000.0f; /* Initialize with high value (indicates failure) */
```
mbarchie: It might be good to make a 10000.0f a named constant instead of using magic number here. Thresholds are #defines so this one could be also for consistency and maintainability.

samples/nrf_edgeai/classification/src/main.c
rrusak: In real application and  sensor streaming scenario it is more convenient to wait positive status than handle negative (IN_PROGRESS or some ERROR), for my point of view

Szynkaa: now other errors are not handled as well, the loop just continues if they are present

it's up to you

samples/nrf_edgeai/anomaly/src/main.c
```diff
@@ -0,0 +1,350 @@
...
+
+	printk("The model correctly distinguishes between healthy and faulty gears.\r\n");
+
+	while (1) {
+		printk("\n========== All test cases completed ==========\r\n");
+		k_sleep(K_MSEC(5000));
+	}
+
+	return 0;
+}
```
mbarchie: No newline at the end of the file

samples/nrf_edgeai/anomaly/src/main.c
```diff
@@ -0,0 +1,350 @@
...
+	/* Validate that the loaded model matches our expected configuration */
+	assert(nrf_edgeai_input_window_size(p_user_model) == USER_WINDOW_SIZE);
+	assert(nrf_edgeai_uniq_inputs_num(p_user_model) == USER_UNIQ_INPUTS_NUM);
+
+	/* Initialize Edge AI runtime for inference execution */
+	nrf_edgeai_err_t res = nrf_edgeai_init(p_user_model);
+	assert(res == NRF_EDGEAI_ERR_SUCCESS);
+
+	flt32_t anomaly_score;
+	size_t DATA_LEN = USER_WINDOW_SIZE * USER_UNIQ_INPUTS_NUM;
```
mbarchie: This value is constant, so a const qualifier could be added to protect against modifications.

include/nrf_edgeai/dsp/private/nrf_dsp_c_intrinsics.h
rrusak: yeah maybe you confused because name of the function if clip_i64_to_i16 and not clip_q63_to_q15...

Here is about Q-format for fixedpoint math: 
https://shafq.at/introduction-to-fixed-point-signal-processing.html#:~:text=Getting%20Started%20with%20Fixed%2Dpoint,What%20happened%20to%20+1.0?

We checks overflow before the final >>15, using the top 32 bits of the 64-bit number. Checking (x >> 16) and (x >> 15) would be too late and would destroy the information needed to reliably detect overflow.
We want compute: (int16_t)(x >> 15) but **only if that result fits into int16_t** (−32768…32767), so it must detect whether:
x >> 15 < -32768   or   x >> 15 > 32767
this means 
x <  -1073741824   or   x >  1073741823 

(in other words (INT16_MIN << 15) and (INT16_MAX << 15))

So the function needs to detect whether the 64-bit number x is outside that safe range before applying the final >>15.

**The test:**
((int32_t)(x >> 32)) != ((int32_t)x >> 31)
compares:
Left side: (int32_t)(x >> 32) → the upper 32 bits of the 64-bit number (bits 63..32)
Right side: ((int32_t)x >> 31) → sign-extend the lower 32 bits, producing either 0x00000000 or 0xFFFFFFFF

**This comparison answers:**
“Are all high bits (above the final 16-bit result) consistent with the sign of the value that would result after >>15?”

If the high 32 bits do not match the sign, overflow would occur.

You detect overflow by checking the upper half of the accumulator.

Because after applying >>15 later, only the top 16 bits matter.
----------------------------------------------------------------------------
**Why NOT check using (x >> 16) vs (x >> 15)?**

you are comparing:
- upper bits after partially shifting down 16
- the lower 32 bits shifted by 15 (sign bit of bit 15)

This fails for three reasons:
**Reason 1 You’ve already destroyed information needed for overflow detection**
Shifting the whole number by 16 bits removes crucial bits that tell you whether the number would overflow after the final ·15 shift.
By the time you detect overflow, the damaging shift has already happened — it's too late to know whether (x >> 15) would exceed ±32767.

This is the same as: Checking after the overflow has already occurred.

**Reason 2  asymmetric effects around zero**

>>15 boundary corresponds to the range:
x in [-32768*32768, 32767*32768 + (2^15 - 1)]

But shifting by 16 moves the boundary:
x in [-32768*65536, 32767*65536 + ...]

This changes the clipping thresholds and makes the check wrong.

**Reason 3 the bit pattern logic breaks completely**

The function relies on this invariant:

_Valid no-overflow range produces:
    bits 63..31 = all 0s (positive) or all 1s (negative)_

But after shifting by 16: 
- bits 63..16 move down
- the invariant is destroyed
- the sign consistency check becomes meaningless

You are no longer checking whether the upper half can hold the sign of the shifted result.

Maybe it is too much text and explanation..... but I think you need to check Q-notation first, then we can have a separate discussion why it is done in such way :D

samples/nrf_edgeai/anomaly/src/main.c
```diff
@@ -0,0 +1,350 @@
...
+		if (res == NRF_EDGEAI_ERR_SUCCESS) {
+			/* Input buffer has reached 128 samples—run inference on the complete window
+			 */
+			res = nrf_edgeai_run_inference(p_user_model);
+
+			/* Check that inference executed successfully */
+			if (res == NRF_EDGEAI_ERR_SUCCESS) {
+				/* Extract anomaly score from model's decoded output */
+				anomaly_score = p_user_model->decoded_output.anomaly.score;
+			}
```
mbarchie, line 261: This part is assumed to be executed only once, but in theory it could be executed multiple times if there are more samples than number required to run inference and user will get the result of inference only for the last run. Maybe in such a case it would be clearer if the loop was broken after first successful inference?

samples/nrf_edgeai/anomaly/src/main.c
```diff
@@ -0,0 +1,350 @@
...
+ * **Practical Application:**
+ * Once validated, this model enables continuous monitoring of rotating machinery,
+ * detecting early signs of gear wear, cracks, misalignment, or bearing failure,
+ * triggering maintenance alerts before catastrophic failures occur.
+ *
+ */
+int main(void)
+{
+	/*  Get user generated model pointer */
+	nrf_edgeai_t *p_user_model = nrf_edgeai_user_model();
```
mbarchie, line 312: You can optionally consider asserting against NULL pointer. I know that nrf_edgeai_user_model() returns pointer to statically allocated object. However, it's an external function, so for the sake of maintainability, there could be a protection against null pointer in case its implementation changes.

samples/nrf_edgeai/anomaly/src/main.c
```diff
@@ -0,0 +1,350 @@
...
+// ////////////////////// Package Group Header Files /////////////////////////
+#include <nrf_edgeai/nrf_edgeai.h>
+#include "nrf_edgeai_generated/nrf_edgeai_user_model.h"
+// /////////////////// Application Global Header Files ///////////////////////
+// /////////////////// 3rd Party Software Header Files ///////////////////////
+#include <zephyr/kernel.h>
+// ////////////////////// Standard C++ Header Files //////////////////////////
+// /////////////////////// Standard C Header Files ///////////////////////////
+#include <stdio.h>
+#include <assert.h>
```
mbarchie: Comment style is not consistent throughout the source file. Zephyr coding style recommends /**/ comments, like it's done anywhere else in this file.

samples/nrf_edgeai/anomaly/Kconfig.sysbuild
rrusak: yes your right, it is not necessary, removed

samples/nrf_edgeai/regression/src/main.c
```diff
@@ -0,0 +1,313 @@
...
+			     [19] = {1.4, 920, 783, 1046, 1550, 588, 24.5, .384, 1.1669, 5.4},
+			     [20] = {1.3, 906, 790, 893, 837, 642, 12.3, 19.3, .274, 5.5},
+			     [21] = {0.6, 882, 563, 978, 936, 660, 8.9, .527, .6034, 1.7},
+			     [22] = {3.4, 1403, 1443, 508, 2234, 1811, 25.2, .386, 1.2215, 25.6},
+			     [23] = {3.9, 1297, 1102, 507, 1375, 1583, 18.2, .363, .7487, 13.6},
+			     [24] = {1.3, 987, 800, 989, 1462, 658, 15.5, .571, .996, 5.7},
+			     [25] = {1.3, 869, 866, 1107, 1212, 596, .238, .145, .4222, 7.2},
+			     [26] = {3.2, 1336, 1340, 540, 2049, 1400, 21.3, .635, 1.5941, 21.6},
+			     [27] = {1, 955, 723, 1129, 1393, 559, 27.7, .258, .9467, 4.2},
+			     [28] = {4.3, 1373, 1364, 597, 2005, 1745, 33.7, .226, 1.1658, 22.5}};
```
mbarchie: 1) Inconsistent format of decimals (some have leading zeroes, some don't)
2) Even though this will probably have internal linkage, the array could be made "static struct..." to express the intent explicitly

samples/nrf_edgeai/regression/src/main.c
```diff
@@ -0,0 +1,313 @@
...
+ * - Returns an invalid sentinel value (-1000.0) if any step fails, allowing the caller
+ *   to detect and handle prediction failures gracefully.
+ * - Assertions validate that the model outputs exactly 1 value as expected for this
+ *   regression task.
+ */
+static flt32_t model_predict(nrf_edgeai_t *p_user_model, flt32_t *p_input_features,
+			     size_t features_num)
+{
+	nrf_edgeai_err_t res;
+	flt32_t model_prediction = -1000.0; /* Invalid default value for error detection */
```
mbarchie: It's an arbitrary magic value, so maybe consider creating a named constant.
Could it also be a valid value of prediction? In such case maybe e.g. NAN value could be used instead?

samples/nrf_edgeai/regression/src/main.c
```diff
@@ -0,0 +1,313 @@
...
+#include <nrf_edgeai/nrf_edgeai.h>
+#include "nrf_edgeai_generated/nrf_edgeai_user_model.h"
+// /////////////////// Application Global Header Files ///////////////////////
+// /////////////////// 3rd Party Software Header Files ///////////////////////
+#include <zephyr/kernel.h>
+// ////////////////////// Standard C++ Header Files //////////////////////////
+// /////////////////////// Standard C Header Files ///////////////////////////
+#include <stdio.h>
+#include <assert.h>
+#include <math.h>
```
mbarchie: Inconsistent comment style

samples/nrf_edgeai/regression/src/main.c
```diff
@@ -0,0 +1,313 @@
...
+	printk("\n--- Testing Model Air Quality predictions ---\r\n");
+	/* Validation loop: test the model against all 29 sample data points */
+	for (size_t i = 0; i < sizeof(USER_INPUT_DATA) / sizeof(USER_INPUT_DATA[0]); i++) {
+		/* Extract sensor readings and environmental parameters from test sample i */
+		flt32_t ground_truth =
+			fill_features_buffer(input_features, USER_UNIQ_INPUTS_NUM, i);
+
+		/* Run neural network inference with the extracted features */
+		flt32_t predicted_value =
+			model_predict(p_user_model, input_features, USER_UNIQ_INPUTS_NUM);
```
mbarchie, line 302: Would it be good to do simple error handling, to check if inference has been run at all?

samples/nrf_edgeai/regression/src/main.c
```diff
@@ -0,0 +1,313 @@
...
+	/* Initialize the Edge AI runtime to prepare the model for inference execution */
+	nrf_edgeai_err_t res = nrf_edgeai_init(p_user_model);
+	assert(res == NRF_EDGEAI_ERR_SUCCESS);
+
+	/* Allocate buffer for holding the 9 input features before each inference */
+	flt32_t input_features[USER_UNIQ_INPUTS_NUM];
+
+	printk("\n--- Testing Model Air Quality predictions ---\r\n");
+	/* Validation loop: test the model against all 29 sample data points */
+	for (size_t i = 0; i < sizeof(USER_INPUT_DATA) / sizeof(USER_INPUT_DATA[0]); i++) {
```
mbarchie: It may not be obvious for some users what is
`sizeof(USER_INPUT_DATA) / sizeof(USER_INPUT_DATA[0])`
supposed to be.

Consider replacing it with a macro, e.g. NUM_ARR_ELEMS() or something like this for better readability.

mbarchie (commented):
I've just started exploring **sdk-edge-ai** repository, so as for now here are my initial comments for **anomaly** and **regression** samples.

Szynkaa (approved):

samples/nrf_edgeai/regression/src/main.c
rrusak: /* Run neural network inference with the extracted features */
		flt32_t predicted_value =
			model_predict(p_user_model, input_features, USER_UNIQ_INPUTS_NUM);

		/* Calculate absolute error: magnitude of difference between prediction and ground
		 * truth */
		flt32_t abs_err = fabsf(predicted_value - ground_truth);
		assert(abs_err <= EXPECTED_MODEL_MAE);
		
Done (if inference is failed function will return INVALID_PREDICTION_VALUE and it will not pass assert )

wbober (approved):

annwoj (approved):

PR #21, opened 2026-01-05
samples: Add hello_ei sample for Edge Impulse

mbarchie:
This pull request proposes a new "Hello Edge Impulse" sample application for Zephyr, demonstrating how to integrate and use the Edge Impulse machine learning library on Nordic Semiconductor devices. The sample includes CMake build integration to fetch and build the Edge Impulse library, configuration options, sample input data, and a main application that runs inferences on example datasets.

The samples replaces `wrapper` sample from nrf repository in a way that it demonstrates running inference with Edge Impulse. However, it is a simplified version, without extensive abstraction layer (`ei_wrapper` library) for interaction with Edge Impulse. Instead, C linkage of EI library is utilized, so no need to introduce any wrapper layer. This means that memory management or time measurements in the new sample are handled by application itself, not external component (library).

Another difference between new sample and `wrapper` samples is existence of 2 input signals compiled-in (sine wave and triangle wave). Both signals are extended with 8 additional elements at the end, to allow multiple inferences.

Jira: NCSDK-36946


samples/edge_impulse/hello_ei/sample.yaml
mbarchie, line 1: Currently this is taken as-is from `wrapper` sample

samples/edge_impulse/hello_ei/prj.conf
```diff
@@ -0,0 +1,26 @@
...
+CONFIG_FP16=n
+CONFIG_REQUIRES_FULL_LIBCPP=y
+CONFIG_CBPRINTF_FP_SUPPORT=y
+CONFIG_FPU=y
+
+# Increase main stack size
+CONFIG_MAIN_STACK_SIZE=2048
+
+# Use URL to fetch Edge Impulse library from the Internet or local path
+CONFIG_EDGE_IMPULSE_URI="edge_impulse/nrf_accel_sim-v35.zip"
```
mbarchie, line 25: EI lib should be updated

samples/edge_impulse/hello_ei/edge_impulse/ei_standalone_lib.cmake
Szynkaa, line 1: I think we should try to simplify this CMake and the template. For example we are no longer downloading zip

samples/edge_impulse/hello_ei/edge_impulse/ei_standalone_lib.cmake
```diff
@@ -0,0 +1,144 @@
...
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/third_party
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/tensorflow
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/dsp
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/classifier
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/anomaly
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/CMSIS/NN/Include
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/CMSIS/DSP/PrivateInclude
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/CMSIS/DSP/Include
+  ${EDGE_IMPULSE_SOURCE_DIR}/edge-impulse-sdk/CMSIS/Core/Include
+)
```
Szynkaa, line 113: As this involve internal structure of EI SDK, I think this should be in template file

samples/hello_ei/src/ei_wrapper.cpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#include "ei_run_classifier.h"
+
+#if !CONFIG_ZTEST
+/* Fixes warnings about redefinition of Zephyr ROUND_UP macro. */
+#ifdef ROUND_UP
+#undef ROUND_UP
+#endif
+#endif /* !CONFIG_ZTEST */
```
Szynkaa: wrapper in no longer tested with ZTEST

samples/hello_ei/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+cmake_minimum_required(VERSION 3.20.0)
+
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(hello_ei)
+
+# Add Edge Impulse library as a separate module
+add_subdirectory(edge_impulse)
```
Szynkaa: minor: move it next to link library

samples/hello_ei/src/main.c
```diff
@@ -0,0 +1,165 @@
+/*
+ * Copyright (c) 2026 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+
+#include "ei_classifier_types.h"
```
Szynkaa: I think we should use more complete path of `<edge-impulse-sdk/classifier/ei_classifier_types.h>`. We already do this for model metadata. Note the angle brackets as EI SDK is build a library.

samples/hello_ei/src/ei_wrapper.cpp
```diff
@@ -0,0 +1,40 @@
...
+int ei_wrapper_run_inference(ei_impulse_result_t *ei_result, size_t window_size)
+{
+	signal_t features_signal;
+
+	__ASSERT_NO_MSG(ei_result != nullptr);
+
+	features_signal.get_data = get_data_cbk;
+	features_signal.total_length = window_size;
+
+	EI_IMPULSE_ERROR err = run_classifier(&features_signal, ei_result, DEBUG_MODE);
```
Szynkaa: In file `edge-impulse-sdk/classifier/ei_run_classifier_c.h` there is already extern "C" function ei_run_classifier. Maybe we can get rid of this wrapper entirely.
Check also https://github.com/edgeimpulse/example-standalone-inferencing-c

samples/edge_impulse/hello_ei/edge_impulse/ei_standalone_lib.cmake
```diff
@@ -0,0 +1,144 @@
...
+add_dependencies(edge_impulse_project zephyr_generated_headers)
+add_library(edge_impulse_imported STATIC IMPORTED)
+set_target_properties(edge_impulse_imported PROPERTIES IMPORTED_LOCATION ${EDGE_IMPULSE_LIBRARY})
+target_link_libraries(edge_impulse_imported INTERFACE kernel)
+
+# This targets remove the `edge_impulse_project-download` stamp file, which
+# causes the Edge impulse library to be fetched on each build invocation.
+# Note: This also results in the `ALL` target to always be considered out-of-date.
+if(CONFIG_EDGE_IMPULSE_DOWNLOAD_ALWAYS)
```
zycz, line 88: Shouldn't we either remove it or add to Kconfig and actually do download?

samples/hello_ei/edge_impulse/CMakeLists.ei.template
```diff
@@ -0,0 +1,81 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor ASA
```
zycz: I would either put year 2026 or move whole commit history of that file and leave old date.

IMO second option is not worth effort. You can mention authors of previous implementation in commit message.

Same for other files.

samples/hello_ei/src/ei_wrapper.cpp
```diff
@@ -0,0 +1,40 @@
...
+#if !CONFIG_ZTEST
+/* Fixes warnings about redefinition of Zephyr ROUND_UP macro. */
+#ifdef ROUND_UP
+#undef ROUND_UP
+#endif
+#endif /* !CONFIG_ZTEST */
+
+#include "ei_wrapper.h"
+
+#define DEBUG_MODE    IS_ENABLED(CONFIG_EI_WRAPPER_DEBUG_MODE)
```
zycz: I would remove this define and use IS_ENABLED() as we use it in only one place.

samples/hello_ei/src/main.c
```diff
@@ -0,0 +1,165 @@
...
+			delta = k_uptime_delta(&start_time);
+
+			if (err != 0) {
+				LOG_ERR("Inference failed with error code: %d", err);
+				return -1;
+			}
+
+			print_inference_result(&inference_result, delta);
+
+			// Keep track of how many inferences have been performed
```
zycz: We are supposed to use /* */ comment style for multi line comments.
Also in other places

samples/hello_ei/src/main.c
```diff
@@ -0,0 +1,165 @@
...
+#include "ei_classifier_types.h"
+#include "model-parameters/model_metadata.h"
+
+#include "ei_wrapper.h"
+#include "input_data.h"
+
+LOG_MODULE_REGISTER(hello_ei);
+
+#define INPUT_WINDOW_SIZE    EI_CLASSIFIER_DSP_INPUT_FRAME_SIZE
+#define LABEL_COUNT          EI_CLASSIFIER_LABEL_COUNT
```
zycz: You define it here and later you still use EI_CLASSIFIER_LABEL_COUNT

zycz: same for EI_CLASSIFIER_DSP_INPUT_FRAME_SIZE

Used in print_model_info(void)

samples/edge_impulse/hello_ei/prj.conf
```diff
@@ -0,0 +1,26 @@
...
+#
+
+CONFIG_NCS_SAMPLES_DEFAULTS=y
+
+# Using console with UART
+CONFIG_CONSOLE=y
+CONFIG_UART_CONSOLE=y
+CONFIG_RTT_CONSOLE=n
+
+# Enable Edge Impulse dependencies
```
zycz, line 13: I wonder if it isn't more neat to have it under Kconfig dependencies/selects.
This would require however config EDGE_IMPULSE so if you think otherwise leave it as it is.

samples/hello_ei/edge_impulse/CMakeLists.txt
```diff
@@ -0,0 +1,144 @@
...
+  edge_impulse_project_download
+)
+
+# Suppress known build warnings for Edge Impulse library header files.
+target_compile_options(edge_impulse
+  INTERFACE -Wno-double-promotion
+  INTERFACE -Wno-unused
+)
+
+if(CONFIG_EI_WRAPPER)
```
zycz: This also needs fixing/removing

samples/hello_ei/sample.yaml
```diff
@@ -0,0 +1,47 @@
+sample:
+  description: Sample showing basic Edge Impulse usage
+  name: Hello Edge Impulse sample
+common:
+  harness: console
+  harness_config:
+    ordered: true
+    regex:
+      - "Classification results"
```
zycz: regex needs to be updated

samples/hello_ei/sample.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - "Value: (1[.]0[0-9]+)|(0[.]9[0-9]+)\tLabel: sine"
+      - "Value: 0[.]0[0-9]+\tLabel: triangle"
+      - "Value: 0[.]0[0-9]+\tLabel: idle"
+      - "Anomaly: (-[0-9]+[.][0-9]+)|(0[.]0[0-9]+)"
+    type: multi_line
+  platform_allow:
+    - nrf52dk/nrf52832
```
zycz: Should we support all SoCs? I guess we can drop nrf52 and maybe some other platforms? This probably should be clarified on meeting with PMs.

mbarchie:
WIP, I'm still working on part of the issues raised in PR

samples/hello_ei/src/main.c
mbarchie: I agree, we can switch to using complete path, then we can reduce number of include directories in cmake files.

I'm not sure about angle bracket though. If we try to simplify cmake system for this sample (compile SDK into app target), this won't be external library anymore. Even if we leave the current approach (SDK as static lib linked against app), the lib still will be local to the sample application, not a system-wide one, so I guess it's acceptable to leave quotes instead of angle brackets.

samples/hello_ei/src/ei_wrapper.cpp
mbarchie: Great point, I managed to get rid off it.

samples/hello_ei/sample.yaml
mbarchie: Yes, I agree, for now it was just copied from the wrapper sample. I'll add this to the question list being prepared.

samples/edge_impulse/hello_ei/prj.conf
mbarchie: I get an idea, but I'll postpone this decision until the way of handling SDK (standalone lib vs in-app) is clarified.

samples/edge_impulse/hello_ei/edge_impulse/ei_standalone_lib.cmake
mbarchie: I think it's not possible. Template file is used for replacing main cmake file in EI SDK. It's used for compiling a static lib and it already keeps track of include directories. These include directories here are needed for the interface library that is used as a "glue" between libedge_impulse.a and the application. This way app knows where to look for headers.

However, the zephyr interface library doesn't need all the paths - just ${EDGE_IMPULSE_SOURCE_DIR} would be enough if the app refers to SDK EI using full paths (e.g. "edge-impulse-sdk/classifier/ei_classifier_types.h").

samples/edge_impulse/hello_ei/edge_impulse/ei_standalone_lib.cmake
mbarchie: I restored it for now

samples/edge_impulse/hello_ei/edge_impulse/ei_standalone_lib.cmake
mbarchie: Yes, that can be simplified for sure, or at least should be documented in a better way.

I'm not sure about download functionality though. I mean, we probably would like to keep SDK deployed as a zip (?). Then we also need to extract this archive and ExternalProject_Add() gives us both things (extract & download) "for free". The zip provided with sample can be a local file, but with a possibility to retrieve it from somewhere else if a user wants it.

I prepared a following commit that documents this file a little bit more, extracts URI handling to separate cmake file.

mbarchie:
The latest commit introduces dual compilation mode - SDK compiled into app vs SDK kept as a static lib.

It doesn't mean I push this as an ultimate solution, it's kept for purpose of comparison of 2 options. Personally, I like the simplicity of a version with SDK compiled into application (without patching, additional interface lib target, exported lib target etc.), but I wonder if there might be some use case or significant benefit of keeping also option with isolated SDK binary.

PR #25, opened 2026-01-19
edge_impulse: data_forwarder: Move sample from NCS

mbarchie:
Migrates the Edge Impulse data forwarder sample from NCS to sdk-edge-ai with improvements.

### Overview

The data forwarder sample enables real-time sensor data collection in Edge Impulse Studio via UART for ML model training. Uses simulated 3-axis accelerometer data with configurable wave patterns.

### Key Changes

**Architecture**
- Timer + work queue pattern: Uses `K_TIMER_DEFINE` and `K_WORK_DEFINE` for periodic sensor sampling (every 100ms) without blocking
- Single `provide_sensor_data()` function handles both sensor reading and UART transmission
- Async UART with atomic busy flag to prevent transmission overlap
- `BUILD_ASSERT` validates sensor channel count matches output format

**Configuration**
- Dual UART setup: Separate UARTs for console logging and data forwarding
  - Data forwarding: UART20
  - Console: UART30
- Common `app.overlay` configures sensor_sim and UART assignments for supported platforms

**Testing**
- Console harness validates CSV output format with regex
- Requires 10 consecutive valid samples
- Supports both positive and negative values

**Platform Support**
- nRF54L15DK (nrf54l15/cpuapp)
- nRF54LM20DK (nrf54lm20a/cpuapp)

Jira: NCSDK-37011


zycz:
We have a custom to link to Jira ticket in commit message. In nrf it would automatically link PR in Jira ticket. It won't work here but still I guess it is good idea IMO.

doc/integrations/edge_impulse.rst
```diff
@@ -0,0 +1,96 @@
...
+Before you start the |EAI| integration with Edge Impulse, make sure that the following prerequisites are completed:
+
+* :ref:`nrf_edgeai_setting_up_sdk`
+* Setup of the required Development Kit (DK).
+* Creation of an `Edge Impulse studio account <Edge Impulse studio signup_>`_ and an Edge Impulse project.
+
+Solution architecture
+*********************
+
+The usage of `Edge Impulse`_ SDK is demonstrated by the Hello Edge Impulse sample.
```
zycz: It is not yet present. Maybe we should add info about it when it will be added? Or even add whole integration section when that sample is present?

samples/edge_impulse/data_forwarder/src/main.c
```diff
@@ -0,0 +1,150 @@
...
+		LOG_ERR("Sensor device not ready");
+		return -ENODEV;
+	}
+
+	if (!device_is_ready(uart_dev)) {
+		LOG_ERR("UART device not ready");
+		return -ENODEV;
+	}
+
+	int err = uart_callback_set(uart_dev, uart_cb, NULL);
```
zycz, line 60: It would be nicer to be consistent and use one way of handling return value.

It is not important and you can ignore it if you prefer.

zycz (approved):

doc/integrations/edge_impulse.rst
mbarchie: I considered this aspect, I just didn't want to remove some parts of the doc just to restore them in hello_ei commit. However, I agree it would be more consistent not to have this reference to unexisting sample. I'll try to make it cleaner.

samples/edge_impulse/data_forwarder/src/main.c
mbarchie: What do you mean by this, what is your suggestion? ``uart_callback_set()`` return value is treated differently than ``*_is_ready()`` calls because it returns actual error code, not boolean.

zycz: ok, you are correct :)

doc/links.txt
```diff
@@ -22,6 +22,7 @@
 .. _`Power Profiler Kit II (PPK2)`: https://docs.nordicsemi.com/bundle/ug_ppk2/page/UG/ppk/PPK_user_guide_Intro.html
 .. _`Power Profiler app`: https://docs.nordicsemi.com/bundle/nrf-connect-ppk/page/index.html
 .. _`nRF Connect for Visual Studio Code: Thread Viewer`: https://docs.nordicsemi.com/bundle/nrf-connect-vscode/page/reference/ui_panel.html#thread-viewer
+.. _`Simulated sensor driver`: https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/drivers/sensor_sim.html
```
Szynkaa: I think it should go to next section. Or be removed and keep just a name in sample's doc

samples/edge_impulse/data_forwarder/src/main.c
```diff
@@ -0,0 +1,150 @@
...
+		return err;
+	}
+
+	/* Send data over UART. */
+	static uint8_t buf[UART_BUF_SIZE];
+
+	if (!atomic_cas(&uart_busy, false, true)) {
+		LOG_ERR("UART not ready - please use lower sampling frequency");
+		return -EBUSY;
+	}
```
Szynkaa, line 95: minor: we can move it down just before accessing uart

samples/edge_impulse/data_forwarder/src/main.c
```diff
@@ -0,0 +1,150 @@
...
+static void sensor_read_work_handler(struct k_work *work)
+{
+	ARG_UNUSED(work);
+	provide_sensor_data();
+}
+
+static void sensor_timer_handler(struct k_timer *timer_id)
+{
+	ARG_UNUSED(timer_id);
+	k_work_submit(&sensor_read_work);
```
Szynkaa: I think we should check if the work was not already submitted as it will result in missing sensor sample. Also we are delaying fetching sensor sample by unknown value depending on number of system works in this queue

doc/links.txt
mbarchie: I'll move it to the next section

annwoj (approved):

samples/edge_impulse/data_forwarder/src/main.c
mbarchie: I moved sensor and UART handling to work item not to do this in ISR context, but I get your point.

What do you think of the following solution?

```
static void sensor_timer_handler(struct k_timer *timer_id)
{
	ARG_UNUSED(timer_id);

	if (k_sem_count_get(&sensor_sem) > 0) {
		LOG_ERR("Previous sensor read still pending - missed sample!");
	}
	k_sem_give(&sensor_sem);
}
```
```
while (1) {
	k_sem_take(&sensor_sem, K_FOREVER);
	provide_sensor_data();
}
```

This seams to simplify things a little bit (no work queue), you have more control on when provide_sensor_data() is executed and also get warnings about lost samples.

Szynkaa: This does not assure that provide_sensor_data has completed, as we take semaphore before the call. That's expected as semaphore provides one-way synchronization. We should probably repurpose the `uart_busy` atomic to signal completion of the call

mbarchie: I agree that we can encounter false negative here, we'll just avoid calling sensor handling before previous call finishes. Your idea is fine or other options probably too:
1) adding a dedicated atomic flag for processing state - `uart_busy` only tracks UART state, not the full processing path (sensor reading + UART TX); a separate flag makes the intent clearer.
2) alternatively, using a second semaphore as a mutex pattern for bi-directional synchronization.

It seems that atomic flag solution is simpler and sufficient for this case.

Szynkaa: That's why I have described it as repurpose. `uart_busy` is set false after full processing path. So if we rename it, and set true after taking semaphore, we will be checking if previous processing completed

samples/edge_impulse/data_forwarder/src/main.c
Szynkaa: sorry for false information, this is also protecting `buf` from access

samples/edge_impulse/data_forwarder/src/main.c
mbarchie: There is a potential case in which we can safely read data from sensors while UART being busy, and UART might become available in the meantime, later, but before we're trying to write to it. In other words, we loose benefit of UART async API.

I know it's kind of edge case and readout in current implementation probably does not take too long, but if this code is replaced with some more time-consuming readout, there might be a benefit of using separate flag.

Szynkaa: OK, feel free to add new atomic

Szynkaa (approved):

grochu (approved):

PR #26, opened 2026-01-20
doc: add selected sphinx extensions

Szynkaa:
Add selected sphinx extensions from zephyr and nrf.

Ref: NCSDK-37180

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -14,6 +14,22 @@
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
 
+import sys
+from pathlib import Path
+
+import west.manifest
+
+manifest = west.manifest.Manifest.from_topdir()
+
+EDGE_AI_BASE = Path(manifest.repo_abspath)
+NRF_BASE = Path(manifest.get_projects(['nrf'])[0].abspath)
+ZEPHYR_BASE = Path(manifest.get_projects(['zephyr'])[0].abspath)
+
+sys.path.insert(0, str(NRF_BASE / 'doc' / '_extensions'))
+sys.path.insert(0, str(ZEPHYR_BASE / 'doc' / '_extensions'))
+
+# Needed by options_from_kconfig extension which is not self contained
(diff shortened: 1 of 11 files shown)


doc/includes/sample_board_rows.txt
```diff
@@ -0,0 +1,51 @@
+.. heading
+
+| Hardware platforms | PCA | Board name | Board target |
+
+
+
+.. nrf52dk_nrf52832
```
zycz, line 7: Should we add all those platforms?

doc/conf.py
```diff
@@ -32,7 +48,10 @@
 # ones.
 extensions = [
     'sphinx_tabs.tabs',
-    'sphinx_copybutton'
+    'sphinx_copybutton',
+    'options_from_kconfig',
+    'table_from_rows',
+    'zephyr.external_content',
```
mbarchie, line 54: A question just to make sure - does this part also make `table-from-sample-yaml` macro available?

zycz (approved):

grochu (approved):
Looks good. You might mention which extensions are added in the commit message

doc/conf.py
Szynkaa: yes

doc/includes/sample_board_rows.txt
Szynkaa: All platforms except Thing:53 are used by anomaly/classification/regression samples. Thingy:53 is going to be used by gesture recognition app

mbarchie (approved):

doc/includes/sample_board_rows.txt
zycz: But shouldn't we drop nrf52 series?

Szynkaa: IMO it can be removed after it's removed from existing samples

rrusak (approved):

annwoj (approved):

PR #27, opened 2026-01-22
Render Axon manually

Szynkaa:
Render Axon NPU driver, platform, compiler and simulator by copying from artifacts repository.

After copy minimal adaptation of Kconfig and CMakeLists was made.

These commits will be squashed before merge


tools/axon_compiler/app/nrf_axon_app_nn_compiler.c
```diff
@@ -0,0 +1,125 @@
...
+#include "nrf_axon_driver.h"
+#include "nrf_axon_nn_infer.h"
+#include "nrf_axon_nn_compiler.h"
+#include "nrf_axon_stringization.h"
+#include "nrf_axon_platform.h"
+
+/*
+* Create the model include header file name and structure from 
+* the model name.
+*/
```
rrusak: Maybe it's worth leaving only compiled binaries/executables with Python scripts to call them, rather than forcing the user to compile something before use

lib/axon/platform/src/simulator/nrf_axon_platform_simulator_linux.c
```diff
@@ -0,0 +1,565 @@
+/*
+ * Copyright (c) 2024, Nordic Semiconductor ASA. All Rights Reserved.
+ *
+ * The information contained herein is confidential property of Nordic Semiconductor ASA.
+ * The use, copying, transfer or disclosure of such information is prohibited except by
+ * express written agreement with Nordic Semiconductor ASA.
+ */
+#include "nrf_axon_platform_simulator.h"
```
rrusak, line 8: IMHO I don't think the simulator is a "tool" because it can't be compiled separately as a tool, but is software part of a driver/library. It's used when you want to test your code on a PC without axon hardware itself, and replace hardware calls.

tools/axon_compiler/app/nrf_axon_app_nn_compiler.c
Szynkaa: Part of compiler source is left as client can extend compiler with CPU operators like sigmoid activation function

dts/bindings/nordic,axons.yaml
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2022 Nordic Semiconductor ASA
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+description: Nordic's Axons
+
```
Szynkaa, line 5: It's added temporally until board support is provided in nrf/zephyr

mbarchie (approved):

wbober (approved):

rrusak (approved):

zycz (approved):

tools/axon_compiler/app/nrf_axon_app_nn_compiler.c
mikeb-nordic: The initial release will not support user added operations, so compiler source files will not be rendered. For the record, the compiler shared library is delivered as a binary so the user does not have to compile it. The static library was also delivered so that  users can and NN operations and rebuild the shared library.

Szynkaa: yes, approach changed, compiler source was removed for now

mikeb-nordic (approved):

grochu (approved):
This rendition looks good as a first step and I think it should be merged keeping in mind, that the contents are far from ready. Issues to address in future:
- license alignment,
- header files including hierarchy - very chaotic at the moment,
- coding style and documentation - at some points when CI is set up, one of the mandatory checks will be the checkpatch script which will generate tons of errors on the current code for many reasons.

PR #29, opened 2026-02-02
doc: documentation restructure

annwoj:
Moved files.
Created placeholders as per agreed structure.
The docs will be edited in the following PRs


doc/index.rst
```diff
@@ -43,8 +43,14 @@ See the following documentation:
    :caption: Subpages:
 
    setting_up_environment.rst
```
zycz: I wonder about order here. For me intuitive order is:

solution_comparison.rst
 quick_start.rst
setting_up_environment.rst

First I have to know which solution to follow than I jump to quick start of that solution which next sends me to setting up environment.

zycz (approved):
Documentation builds and looks fine.
I would merge it soon and use it as a base for further development of documentation.

doc/index.rst
annwoj: Agree, must've missed it, it's fixed now

PR #37, opened 2026-02-13
doc: add operators page

annwoj:
added operators page with model recommendations

### DIFF
--- a/doc/supported_operators.rst
+++ b/doc/supported_operators.rst
@@ -1,0 +1,305 @@
+.. _supported_operators:
+
+Supported operators
+###################
+
+.. contents::
+   :local:
+   :depth: 2
+
+This section describes the model structure constraints and the set of operators supported by the Axon Compiler.
+
+Model structure
+***************
+
+The Axon Compiler has the following constraints on model structure:
+
+* Supports 8-bit quantized input and output for all layers, with an option to use int32 model output with a configurable radix.
+* Supports stateful behavior between inferences when declared using VarHandle, ReadVariable, or AssignVariable.
+* Allows a maximum of two inputs per node.
+* Supports maximum tensor sizes: 
+  
+  * Height: 1024
+  * Width: 1024
+  * Channels: 1024
+
+* Provides native activation functions: ReLU, ReLU6, and LeakyReLU.
+* Provides CPU activation functions: Sigmoid, Tanh, and Softmax.
+
+.. _supported_operators_reshape:
+
+Memory organization and reshape
+*******************************
+
+Axon stores tensors in memory using the following layout:
+
+.. code-block:: c
+  
+  int8_t my_axon_tensor[channel_count][height][width];
+
+This differs from TFLite, where channels are the innermost dimension:
+
+.. code-block:: c
+  
+  int8_t my_tflite_tensor[height][width][channel_count];
+
+In addition, the Axon hardware aligns the start of each output row to a 32‑bit boundary. 
+When the tensor width is not a multiple of four, this alignment introduces padding bytes at the end of each row.
+Because of these differences, a reshape operation in TFLite, which only changes tensor dimensions without moving data, may require actual data movement on Axon. 
+In such cases, the reshape is executed on the CPU to perform the required memory reorganization.
+
+Many reshape operations are transparent to Axon and do not require data movement, including:
+
+* Dropping or adding an axis when transitioning between 1D and 2D operators. 
+* Flattening a multi‑channel output before a fully connected operator. 
+  The Axon compiler reorders the weights to account for the channel layout.
+* Applying a height/width transpose around an operation (for example, rotating the input by 90 degrees, performing the operation, and then rotating the output back). 
+  In this case, the Axon compiler removes the reshape operations and executes the operation in the unrotated orientation.
+
+.. _supported_operators_list:
+
+Operators
+*********
+
+Operators can run entirely on Axon, entirely on the CPU, or be split between the two.
+Operators executed on the CPU consume CPU cycles and add a small amount of interrupt latency when they appear in the middle of the model, rather than at the beginning or end.
+The following operators are supported in the current version of Axon Compiler:
+
+Convolution operators
+=====================
+
+.. list-table::
+   :header-rows: 1
+
+   * - Operator
+     - Notes / limitations
+     - Target
+     - Compiler version
+   * - Conv1D
+     - | Dilation not supported.
+       | Implemented as a 2D convolution with either height or width = 1.
+       | Max filter width: 32 (when height is 1)
+       | Max filter height: 16 (when width is 1)
+       | Max stride: 31
+     - Axon NPU
+     - 1.0.0
+   * - Depthwise Conv1D
+     - | Max filter width: 32
+       | Max filter height: 16
+       | Max stride: 31
+     - Axon NPU
+     - 1.0.0
+   * - Conv2D
+     - | Dilation supported with width dilation <= 31, and output width 1.
+       | Max filter dimensions: 16 x 16
+       | Max stride: 31
+     - Axon NPU
+     - 1.0.0
+   * - Depthwise Conv2D
+     - | Channel multipliers not supported
+       | Dilation not supported
+       | Max filter dimensions: 16 x 16
+       | Max stride: 31
+     - Axon NPU
+     - 1.0.0
+
+Fully connected layer
+=====================
+
+.. list-table::
+   :header-rows: 1
+
+   * - Operator
+     - Notes / limitations
+     - Target
+     - Compiler version
+   * - Fully Connected
+     - | Maximum input vector length: 2048
+       | Maximum number of neurons: 2048
+     - Axon NPU
+     - 1.0.0
+
+Pooling operators
+=================
+
+.. list-table::
+   :header-rows: 1
+
+   * - Operator
+     - Notes / limitations
+     - Target
+     - Compiler version
+   * - Average Pooling
+     - | No padding
+       | Max filter dimensions: 32 x 32
+     - Axon NPU
+     - 1.0.0
+   * - Max Pooling
+     - | Max filter dimensions: 32 x 32
+       | Maximum input/output size: 1024
+     - Axon NPU
+     - 1.0.0
+   * - Mean
+     - Includes global average pooling functionality
+     - Axon NPU
+     - 1.0.0
+   * - Global Average Pooling
+     - Implemented internally as Mean
+     - Axon NPU
+     - 1.0.0
+
+Activation functions
+====================
+
+.. list-table::
+   :header-rows: 1
+
+   * - Operator
+     - Notes / limitations
+     - Target
+     - Compiler version
+   * - ReLU
+     - Native activation function
+     - Axon NPU
+     - 1.0.0
+   * - ReLU6
+     - Native activation function
+     - Axon NPU
+     - 1.0.0
+   * - LeakyReLU
+     - Native activation function
+     - Axon NPU
+     - 1.0.0
+   * - Sigmoid
+     - Executed on the CPU
+     - CPU
+     - 1.0.0
+   * - Tanh
+     - Executed on the CPU
+     - CPU
+     - 1.0.0
+   * - Softmax
+     - Partially accelerated on Axon NPU, with CPU assistance
+     - Axon NPU | CPU
+     - 1.0.0
+
+Elementwise operators
+=====================
+
+.. list-table::
+   :header-rows: 1
+
+   * - Operator
+     - Notes / limitations
+     - Target
+     - Compiler version
+   * - Add
+     - Vector operation with broadcast on height and/or width
+     - Axon NPU
+     - 1.0.0
+   * - Multiply
+     - Vector operation with broadcast on height and/or width
+     - Axon NPU
+     - 1.0.0
+
+Tensor manipulation operators
+=============================
+
+.. list-table::
+   :header-rows: 1
+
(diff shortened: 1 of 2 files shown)


doc/supported_operators.rst
```diff
@@ -0,0 +1,102 @@
...
+   * - Depthwise Conv2D
+     - No channel multipliers, max filter dimensions 16x16, max stride 31
+     - Axon NPU|CPU
+     - 1.0.0
+   * - Fully connected
+     - Max input length 2048
+     - Axon NPU|CPU
+     - 1.0.0
+   * - Add
+     - Vector operation with broadcast on height and/or width
```
mikeb-nordic: ...and/or channels

doc/supported_operators.rst
```diff
@@ -0,0 +1,102 @@
...
+
+The Axon Compiler supports a wide range of operators commonly used in machine learning models. 
+The supported operators include:
+
+* Supporting 8-bit quantized input and output for all layers, with an option to use int32 model output with a configurable radix.
+* Supporting stateful behavior between inferences when declared using VarHandle, ReadVariable, or AssignVariable.
+* Allowing a maximum of two inputs per node.
+* Supporting maximum tensor sizes (height, width, channels) of 1024, 1024, and 512.
+* Providing native activation functions: ReLU, ReLU6, and LeakyReLU.
+* Providing CPU activation functions: Sigmoid and Tanh.
```
mikeb-nordic: I would distinguish between model architecture/attributes and specific operators.
The bullet points in this section are not operators; they reflect on the restrictions on the types of models Axon supports.
* 8bit quantized models (not float32, not 16bit quantized).
* "allowing a maximum of 2 inputs per node" is also a restriction.
* Maximum tensor sizes are operator specific.
* Native hardware implementation of fused activation functions ReLU, ReLU6, and LeakyReLU. (FYI, "fused" means that the activation function is part of another operation, rather than a stand-alone operation. The difference is that a fused activation function happens on the operator output before quantization, rather than require the output to be dequantized, apply the activation function, re-quantized).
* CPU implementation of Sigmoid and Tanh as discrete operators.

doc/supported_operators.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 40 20 20
+
+   * - Operator
+     - Notes
+     - Target
+     - Compiler version
+   * - Conv1D
```
mikeb-nordic, line 78: We need to review this. TFLite maps a 1D convolution to a 2D convolution where either height or width is 1.  So maybe the documentation should include:
"Implemented as a 2D Convolution with either height or width = 1." 
The restrictions would then make sense that maximum filter width = 32 when input height is 1, and maximum filter height is 16 when input width is 1.

doc/supported_operators.rst
```diff
@@ -0,0 +1,102 @@
...
+   * - Mean
+     - Includes global average pooling
+     - Axon NPU|CPU
+     - 1.0.0
+   * - Global average pooling
+     - Implemented as Mean
+     - Axon NPU|CPU
+     - 1.0.0
+   * - Strided slice
+     - Max stride 31
```
mikeb-nordic: Maximum width stride is 31; strides on height and channels are not limited.

rrusak (approved):

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+
+Model structure
+***************
+
+The Axon Compiler has the following constraints on model structure:
+
+* Supports 8-bit quantized input and output for all layers, with an option to use int32 model output with a configurable radix.
+* Supports stateful behavior between inferences when declared using VarHandle, ReadVariable, or AssignVariable.
+* Allows a maximum of two inputs per node.
+* Supports maximum tensor sizes (height, width, channels) of 1024, 1024, and 512.
```
mikeb-nordic: Channels maximum is now 1024.

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+
+   * - Operator
+     - Notes / limitations
+     - Target
+     - Compiler version
+   * - Conv1D
+     - | Max filter width: 32
+       | Max filter height: 16
+       | Max stride: 31
+     - Axon NPU | CPU
```
mikeb-nordic: It's not clear to me what the intention of the "target" column is. If it indicates where the operator executes, the value should be "CPU" for sigmoid, tanh, and reshape, "AXON + CPU" for softmax, and "AXON" everywhere else.

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+* Supports 8-bit quantized input and output for all layers, with an option to use int32 model output with a configurable radix.
+* Supports stateful behavior between inferences when declared using VarHandle, ReadVariable, or AssignVariable.
+* Allows a maximum of two inputs per node.
+* Supports maximum tensor sizes (height, width, channels) of 1024, 1024, and 512.
+* Provides native activation functions: ReLU, ReLU6, and LeakyReLU.
+* Provides CPU activation functions: Sigmoid, Tanh, and Softmax.
+
+Operators
+*********
+
```
mikeb-nordic, line 63: Can we add words to the effect of:
_Operators can be implemented exclusively in axon, exclusively in the CPU, or a combination of the two. CPU operators will consume some CPU cycles, and also incur a minor additional interrupt latency if the operator is neither the 1st nor the last in the model. _

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+       | Max stride: 31
+     - Axon NPU | CPU
+     - 1.0.0
+   * - Conv2D
+     - | Max filter dimensions: 16 x 16
+       | Max stride: 31
+     - Axon NPU | CPU
+     - 1.0.0
+   * - Depthwise Conv2D
+     - | No channel multipliers supported
```
mikeb-nordic: Also "Dilation not supported".

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+       | Max stride: 31
+     - Axon NPU | CPU
+     - 1.0.0
+   * - Depthwise Conv1D
+     - | Max filter width: 32
+       | Max filter height: 16
+       | Max stride: 31
+     - Axon NPU | CPU
+     - 1.0.0
+   * - Conv2D
```
mikeb-nordic, line 92: Dilation supported in one special case: width dilation <= 31, and output width = 1.

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+
+.. list-table::
+   :header-rows: 1
+
+   * - Operator
+     - Notes / limitations
+     - Target
+     - Compiler version
+   * - Conv1D
+     - | Max filter width: 32
```
mikeb-nordic: Dilation supported in one special case: width dilation <= 31, and output width = 1.

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+   :header-rows: 1
+
+   * - Operator
+     - Notes / limitations
+     - Target
+     - Compiler version
+   * - Average Pooling
+     - | No padding
+       | Max filter dimensions: 32 x 32
+       | Minimum kernel size: 2
```
mikeb-nordic: Remove "minimum kernel size" and "maximum input/output size" for average and max pooling.. The latter is covered earlier by "maximum tensor sizes".

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+   * - Concatenate
+     - No additional limitations specified
+     - Axon NPU | CPU
+     - 1.0.0
+   * - splitV
+     - No additional limitations specified
+     - Axon NPU | CPU
+     - 1.0.0
+
+Limitations and unsupported operators
```
mikeb-nordic: Please remove this section. We don't need to provide an exhaustive list of what is not supported; that is implied by anything not listed in the "supported" sections.

Also, dilation limits are conveyed in the convolution section.

Reshape is also discussed separately.

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+
+The Axon Compiler has the following constraints on model structure:
+
+* Supports 8-bit quantized input and output for all layers, with an option to use int32 model output with a configurable radix.
+* Supports stateful behavior between inferences when declared using VarHandle, ReadVariable, or AssignVariable.
+* Allows a maximum of two inputs per node.
+* Supports maximum tensor sizes (height, width, channels) of 1024, 1024, and 512.
+* Provides native activation functions: ReLU, ReLU6, and LeakyReLU.
+* Provides CPU activation functions: Sigmoid, Tanh, and Softmax.
+
```
mikeb-nordic, line 28: Requesting a new section:

Memory Organization and Reshape
*************************************

Tensors in Axon are organized in memory as 
int8_t my_axon_tensor[channel_count][height][width]. 
This is in contrast to TFLite where channels are the inner most dimension:
int8_t my_tflite_tensor[height][width][channel_count]. 

In addition, Axon hardware writes the start of every output row on a 32bit boundary, resulting in padding bytes at the end of each row if the width is not a multiple of 4.

For these reasons, reshape in TFLite which simply changes the dimensions of the shape without moving any data may require actual data movement in Axon. In these cases an CPU reshape operation is used to perform the data movement. 
Many reshape operations are transparent to Axon:
* Dropping/adding an axis when transitioning between 2D and 1D operators.
* Flattening a multi-channel output before a fully connected operator (Axon compiler will re-order the weights to account for channel ordering).
* Height/width transpose, operation, height/width transpose. (ie, input is rotated 90 degrees, the operation is performed, then the output is unrotated the 90degrees). Axon compiler will drop the reshapes and perform the operation unrotated.

doc/supported_operators.rst
```diff
@@ -0,0 +1,265 @@
...
+     - 1.0.0
+   * - Concatenate
+     - No additional limitations specified
+     - Axon NPU | CPU
+     - 1.0.0
+   * - splitV
+     - No additional limitations specified
+     - Axon NPU | CPU
+     - 1.0.0
+
```
mikeb-nordic: You can add an entry for reshape here, and direct the reader to the memory organization and reshape section for details.

mikeb-nordic (changes requested):

doc/supported_operators.rst
annwoj: added, please check

doc/supported_operators.rst
```diff
@@ -0,0 +1,274 @@
...
+     - Compiler version
+   * - Conv1D
+     - | ilation supported only for width <= 31, and output width 1.
+       | Max filter width: 32
+       | Max filter height: 16
+       | Max stride: 31
+     - Axon NPU | CPU
+     - 1.0.0
+   * - Depthwise Conv1D
+     - | Max filter width: 32
```
mikeb-nordic, line 87: Similar to Conv1D (except dilation isn't supported).
Implemented as a 2D convolution with either height or width = 1.
Max filter width 32 (when height is 1)
Max filter height 16 (when width is 1)
Dilation not supported.

doc/supported_operators.rst
```diff
@@ -0,0 +1,274 @@
...
+       | Max filter height: 16
+       | Max stride: 31
+     - Axon NPU | CPU
+     - 1.0.0
+   * - Conv2D
+     - | Dilation supported only for width <= 31, and output width 1.
```
mikeb-nordic: "Dilation supported with width dilation <= 31, and output width 1."

doc/supported_operators.rst
```diff
@@ -0,0 +1,274 @@
...
+   * - Concatenate
+     - No additional limitations specified
+     - Axon NPU | CPU
+     - 1.0.0
+   * - splitV
+     - No additional limitations specified
+     - Axon NPU | CPU
+     - 1.0.0
+   * - Reshape
+     - | Supported only between convolutional and dense layers, with some limitations on patterns.
```
mikeb-nordic: Remove 
"Supported only between convolutional and dense layers, with some limitations on patterns." 

Replace with
"Performed as a CPU operation in those cases where it is not transparent."

doc/supported_operators.rst
```diff
@@ -0,0 +1,274 @@
...
+  * LeakyReLU
+
+Architectural constraints
+=========================
+
+To stay within compiler and hardware limits, ensure that the model structure adheres to the following constraints:
+
+* Limiting tensor dimensions:
+
+  * Spatial dimensions (height and width) to 1024 or less
```
mikeb-nordic: This is restating the limitations from above, except the channel count limit is 1024, not 512. I think this could simply say "Do not exceed the operator limits listed above."

doc/supported_operators.rst
```diff
@@ -0,0 +1,274 @@
...
+  * Dilation, unless the output dimension is 1 x 1
+  * Depthwise convolution with dilation
+
+Quantization guidance
+=====================
+
+To achieve optimal performance and predictable accuracy:
+
+* Train models using 8‑bit quantization awareness when targeting deployment on Axon.
+* If Sigmoid or Tanh activations are required, consider training with piecewise linear approximations to minimize accuracy degradation.
```
mikeb-nordic: Remove this statement. Piecewise linear approximation is not supported in the toolchain.

miha-nordic:
@mikeb-nordic @annwoj Should that be merged to 2.0.0 release?

annwoj:
> @mikeb-nordic @annwoj Should that be merged to 2.0.0 release?

yes, it should

mikeb-nordic (approved):

PR #39, opened 2026-02-16
Issue37615 gesture recognition gatt

zycz:
applications: gesture_recognition: Implement BLE mode selection
- Added desktop demo application for presenting gesture recognition
- Refactored BLE source files to support multiple modes: HID, NUS, and GATT.
- Enhanced Kconfig to allow users to choose between different BLE modes.
- Added new GATT service implementation for sending inference results.
  GATT service module was implemented by Raman Rusak.
- Updated main application logic to handle data collection and inference based on selected BLE mode.

PR #75, opened 2026-03-30
applications: gesture_recognition: refactor ble_hid to use HID Service from sdk-nrf

grochu:
It's been agreed offline, that the application would benefit from using the ready Bluetooth HID Service module from the sdk-nrf repo, which will hide Zephyr Bluetooth API changes (migration will be handled on sdk-nrf level) and ensure quality of the service implementation. At the same time the HID Service API is quite stable, so it shouldn't generate too much maintenance in the add-on.

**Not addressed in this PR**:

- [ ] Restore authentication setting selectively for HIDs, previously removed by https://github.com/nrfconnect/sdk-edge-ai/pull/39

### DIFF
--- a/applications/gesture_recognition/README.rst
+++ b/applications/gesture_recognition/README.rst
@@ -304,13 +304,13 @@ Testing
       :class: highlight
 
       Predicted class: DOUBLE SHAKE, with probability 96 %
-      BLE HID Key 8 sent successfully
+      BLE HID key sent successfully
       Predicted class: SWIPE RIGHT, with probability 99 %
-      BLE HID Key 32 sent successfully
+      BLE HID key sent successfully
       Predicted class: SWIPE LEFT, with probability 99 %
-      BLE HID Key 16 sent successfully
+      BLE HID key sent successfully
       Predicted class: ROTATION RIGHT, with probability 93 %
-      BLE HID Key 1 sent successfully
+      BLE HID key sent successfully
 
    Once the device is running, Bluetooth LE advertising starts as a HID device and waits for a connection request from the PC.
    Devices can be connected in the same way as a regular Bluetooth keyboard.
@@ -322,9 +322,6 @@ Testing
 
       Connected 9C:B6:D0:C0:CE:FC (public)
       Security changed: 9C:B6:D0:C0:CE:FC (public) level 2
-      Input CCCD enabled
-      Input attribute handle: 0
-      Consumer CCCD enabled
 
    After Bluetooth connection, the device changes LED indication from red to green, or red to blue depending on the keyboard control mode.
    You can now use the device to control media playback or presentation slides by making gestures.
--- a/applications/gesture_recognition/README.rst
+++ b/applications/gesture_recognition/README.rst
@@ -491,6 +491,7 @@ Dependencies
 This sample uses the following |NCS| services:
 
 * `Nordic UART Service (NUS)`_
+* `HID Service`_
 
 This sample uses the following Zephyr libraries:
 
--- a/applications/gesture_recognition/src/ble/hid/ble_hid.c
+++ b/applications/gesture_recognition/src/ble/hid/ble_hid.c
@@ -14,7 +14,7 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/settings/settings.h>
 
-LOG_MODULE_REGISTER(ble_hid, CONFIG_BT_HIDS_LOG_LEVEL);
+LOG_MODULE_REGISTER(ble_hid);
 
 
 #define KEY_ARROW_LEFT (0x50)
--- a/applications/gesture_recognition/src/ble/hid/ble_hid.c
+++ b/applications/gesture_recognition/src/ble/hid/ble_hid.c
@@ -61,6 +61,8 @@ static const struct bt_data ad[] = {
 	BT_DATA_BYTES(BT_DATA_UUID16_ALL,
 				  BT_UUID_16_ENCODE(BT_UUID_HIDS_VAL),
 				  BT_UUID_16_ENCODE(BT_UUID_BAS_VAL)),
+	BT_DATA_BYTES(BT_DATA_GAP_APPEARANCE,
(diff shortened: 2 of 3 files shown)


applications/gesture_recognition/src/ble/hid/ble_hid.c
mbarchie: Should we start advertising if init fails?

applications/gesture_recognition/src/ble/hid/ble_hid.c
```diff
@@ -14,83 +14,55 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/settings/settings.h>
 
-LOG_MODULE_REGISTER(ble_hid, CONFIG_BT_HIDS_LOG_LEVEL);
+#include <bluetooth/services/hids.h>
 
+LOG_MODULE_REGISTER(ble_hid, LOG_LEVEL_INF);
```
Szynkaa:
```suggestion
LOG_MODULE_REGISTER(ble_hid);
```
To depend on `CONFIG_LOG_DEFAULT_LEVEL`

applications/gesture_recognition/src/ble/hid/ble_hid.c
```diff
@@ -373,73 +293,94 @@ int ble_hid_init(ble_connection_cb_t cb)
...
-	if (!ble_connected || !ccc_enabled) {
-		return -1;
+	if (!ble_connected) {
+		return -ENOTCONN;
 	}
 
 	switch (key) {
 	case BLE_HID_KEY_ARROW_LEFT:
-		report[2] = KEY_ARROW_LEFT;
+		kbd_report[2] = KEY_ARROW_LEFT;
```
mbarchie: Could "2" be more self-explanatory? (index `#define` or struct with report layout?)

mbarchie (approved):

applications/gesture_recognition/src/ble/hid/ble_hid.c
```diff
@@ -333,6 +219,38 @@ BT_CONN_CB_DEFINE(conn_callbacks) = {
 	.security_changed = security_changed,
 };
 
+static void hids_service_init(void)
+{
+	int err;
+	struct bt_hids_init_param hids_init_param = {0};
+	struct bt_hids_inp_rep *kbd_rep;
```
Szynkaa: Some of these abbreviations are more cryptic than other. I don't know if it standard in context of HID, but I couldn't decode `kbd` without external check

applications/gesture_recognition/src/ble/hid/ble_hid.c
Szynkaa, line 85: They are also `#define`d so maybe use these macros?

applications/gesture_recognition/src/ble/hid/ble_hid.c
```diff
@@ -346,6 +264,8 @@ static void bt_ready(int err)
 		settings_load();
 	}
 
+	hids_service_init();
```
zycz: I am not sure if I understand it correctly but this probably should go before settings load so that settings have structures where they can be restored to

applications/gesture_recognition/src/ble/hid/ble_hid.c
```diff
@@ -289,8 +169,13 @@ static void disconnected(struct bt_conn *conn, uint8_t reason)
 
 	LOG_INF("Disconnected from %s (reason 0x%02x)", addr, reason);
 
+	err = bt_hids_disconnected(&hids_obj, conn);
```
zycz, line 177: bt_hids_disconnected returns int but err is uint8

applications/gesture_recognition/README.rst
zycz: this is on longer valid

applications/gesture_recognition/src/ble/hid/ble_hid.c
grochu: We can, the consequence will be that the connected device won't see the whole HID service in our attribute table. Perhaps we can advertise without the HID UUID in the advertising data? Or should the application assert in that case?

grochu: Not exactly - the report map follows the USB HID descriptor format and the mentioned line for example defines byte 0 of the keyboard input report array ("Report Size" is in bits). So we don't have it `#define`d.

applications/gesture_recognition/src/ble/hid/ble_hid.c
grochu: Good catch! Not here exactly here but in the `connected()` function - the `err` parameter was reused with implicit cast - fixed.

applications/gesture_recognition/src/ble/hid/ble_hid.c
grochu: According to https://docs.zephyrproject.org/latest/doxygen/html/group__bt__gatt__server.html#gab4d9cfea04e44445d5dc30ad52842b64 `bt_gatt_service_register` should be actually called either before `bt_init` or after `settings_load`.

applications/gesture_recognition/src/ble/hid/ble_hid.c
grochu: Fixed

applications/gesture_recognition/src/ble/hid/ble_hid.c
grochu: renamed

applications/gesture_recognition/README.rst
grochu: fixed

Szynkaa (approved):

mbarchie (approved):

applications/gesture_recognition/src/ble/hid/ble_hid.c
grochu: So in fact it needs to be moved before `bt_init()` - I'll update it.

grochu: Done

zycz (approved):

mbarchie (approved):

PR #83, opened 2026-04-24
Model observability module

mbarchie:
## Summary

Adds an Edge AI observability library that accumulates inference statistics and ships them to Memfault as CBOR Custom Data Recordings.

## Design

The library is split into two compilation layers so the portable core can be unit-tested on a host without Zephyr or ZCBOR.

### Portable core (`nrf_edgeai_obsv_core.h` / `nrf_edgeai_obsv_core.c`)

- `nrf_edgeai_obsv_core_t` holds model metadata and a singly-linked list of registered metrics.
- `nrf_edgeai_obsv_core_update()` fans out each inference result to all metrics via the `nrf_edgeai_obsv_metric_t` vtable.
- `nrf_edgeai_obsv_core_for_each_metric()` drives the snapshot/encode path via a caller-supplied callback.
- Two built-in metrics, Kconfig-gated:
  - **Probability distribution** - per-class histogram with configurable bin edges.
  - **Transition matrix** - class-to-class transitions across consecutive inferences.

  Both use per-instance storage allocated by define macros (`NRF_EDGEAI_OBSV_METRIC_PD_DEFINE` / `_TM_DEFINE`), so multiple independent instances can coexist in one application.
- Custom metrics plug in via the `nrf_edgeai_obsv_metric_t` vtable. `NRF_EDGEAI_OBSV_ENCODE_METRIC_SIZE` and `CONFIG_NRF_EDGEAI_OBSV_EXTRA_ENCODE_BYTES` provide the buffer-sizing hook for custom metric authors.

### Zephyr wrapper (`nrf_edgeai_obsv.h` / `nrf_edgeai_obsv.c`)

- `nrf_edgeai_obsv_ctx_t` wraps the core with a `k_mutex`, serializing `update()`, `encode()`, and registration across threads.

### CBOR encoder (`nrf_edgeai_obsv_encode.c`, `NRF_EDGEAI_OBSV_ENCODE`)

Requires ZCBOR. `nrf_edgeai_obsv_encode_cbor()` serializes model info and metric snapshots per the schema in `obsv.cddl`. Metric `counts` are read directly from per-instance storage under the context lock - no shadow copy is needed.

### Memfault CDR transport (`lib/nrf_edgeai_obsv_memfault/`)

- `nrf_edgeai_obsv_memfault_init()` registers a `sMemfaultCdrSourceImpl` and binds one or more `nrf_edgeai_obsv_ctx_t` instances (up to `CONFIG_NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS`).
- `nrf_edgeai_obsv_memfault_collect()` encodes all contexts into a staging buffer as a CBOR array. Each call overwrites the previous snapshot so the transport always serves the freshest data.
- Optional auto-collect via `k_work_delayable` (`NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT`).

## Tests

- `tests/obsv` (native_sim) - registration/reset state machine, metric iteration, per-metric snapshot correctness, custom metric support.
- `tests/obsv_memfault` (qemu_cortex_m3) - CDR lifecycle, collect error paths, and a **CDDL conformance check**: `obsv.cddl` is compiled to a zcbor decoder at configure time and run against live encoder output, so the wire format cannot silently drift.


lib/nrf_edgeai_obsv/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+zephyr_library()
+
+# Expose this directory so sibling libraries (e.g. nrf_edgeai_obsv_memfault)
+# can reach private headers such as nrf_edgeai_obsv_encode.h.
+zephyr_include_directories(${CMAKE_CURRENT_LIST_DIR})
```
Szynkaa: maybe replace it with relative includes in nrf_edgeai_obsv_memfault.c?

lib/nrf_edgeai_obsv/Kconfig
```diff
@@ -0,0 +1,56 @@
...
+
+if NRF_EDGEAI_OBSV
+
+config NRF_EDGEAI_OBSV_ENCODE
+	bool "CBOR observability encoder"
+	default y
+	depends on ZCBOR
+	help
+	  Builds nrf_edgeai_obsv_encode.c and enables nrf_edgeai_obsv_encode()
+	  in the Zephyr wrapper. Required by the Memfault transport.
```
Szynkaa:
```suggestion
	  in the Zephyr wrapper.
```
this is stated in other Kconfig by `depends on`

lib/nrf_edgeai_obsv/Kconfig
```diff
@@ -0,0 +1,56 @@
...
+	help
+	  Enable the Edge AI observability library.
+	  Collects runtime metrics from class probabilities and exposes them
+	  as typed snapshots that transport layers can serialize as they wish.
+
+if NRF_EDGEAI_OBSV
+
+config NRF_EDGEAI_OBSV_ENCODE
+	bool "CBOR observability encoder"
+	default y
```
Szynkaa: imo default n

Szynkaa: also in other symbols

include/nrf_edgeai/obsv/nrf_edgeai_obsv.h
```diff
@@ -0,0 +1,126 @@
...
+	/** @brief Portable observability state (shared with encode/snapshot paths). */
+	nrf_edgeai_obsv_state_t state;
+} nrf_edgeai_obsv_ctx_t;
+
+/**
+ * @brief Initialize an observability context.
+ *
+ * @param ctx   Context to initialize.
+ * @param model Model metadata to embed in every encoded snapshot.
+ * @return 0 on success, -EINVAL if any argument is NULL or
```
Szynkaa: use `@retval` when having 3 options listed

include/nrf_edgeai_obsv/nrf_edgeai_obsv.h
```diff
@@ -0,0 +1,126 @@
...
+
+#include <zephyr/kernel.h>
+
+#include <nrf_edgeai/obsv/nrf_edgeai_obsv_metrics.h>
+#include <nrf_edgeai/obsv/nrf_edgeai_obsv_state.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
Szynkaa, line 28: maybe add `@defgroup` for doxygen grouping of the API

include/nrf_edgeai/obsv/nrf_edgeai_obsv_encode_sizes.h
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (c) 2026 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef NRF_EDGEAI_OBSV_ENCODE_SIZES_H
+#define NRF_EDGEAI_OBSV_ENCODE_SIZES_H
+
```
Szynkaa: does this file need to be in public includes?

include/nrf_edgeai_obsv/nrf_edgeai_obsv_memfault.h
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @brief Encodes observability metrics as CBOR and stages them for Memfault.
+ *
+ * The CBOR blob is stored internally and handed out by the registered CDR
+ * source on the next transport drain cycle (BT MDS or HTTP periodic upload).
+ *
+ * @return 0 on success, negative errno on failure.
+ */
+int nrf_edgeai_obsv_memfault_collect(void);
```
Szynkaa, line 59: If this can be blocked by the ctx->mutex, please, add a note

include/nrf_edgeai/obsv/nrf_edgeai_obsv_metrics.h
```diff
@@ -0,0 +1,148 @@
...
+	 */
+	void (*update)(const float *p_probs, uint8_t n);
+
+	/**
+	 * @brief Resets accumulated counters without touching configuration.
+	 *
+	 * Called by nrf_edgeai_obsv_reset() to zero counters while preserving
+	 * any configuration set at registration time (e.g. custom bin edges).
+	 * If NULL, the reset falls back to calling init(NULL), which loses
+	 * custom configuration.
```
Szynkaa: maybe just no-op?

include/nrf_edgeai/obsv/nrf_edgeai_obsv_metrics.h
```diff
@@ -0,0 +1,148 @@
...
+ * Pass to nrf_edgeai_obsv_state_register() as the p_cfg argument.
+ * NULL p_cfg gives uniform bins over [0, 1].
+ */
+typedef struct {
+	/**
+	 * Array of (CONFIG_NRF_EDGEAI_OBSV_PROBS_DISTRIBUTION_BIN_NUM + 1) edge values
+	 * in ascending order over [0, 1], e.g. {0.0, 0.25, 0.5, 0.75, 1.0}
+	 * for 4 uniform bins. NULL means uniform bins over [0, 1].
+	 */
+	const float *bin_edges;
```
Szynkaa:
```suggestion
	float bin_edges[CONFIG_NRF_EDGEAI_OBSV_PROBS_DISTRIBUTION_BIN_NUM + 1];
```

include/nrf_edgeai/obsv/nrf_edgeai_obsv_encode_sizes.h
Szynkaa: can some of this defines be extracted from `sizeof` some structs?

Szynkaa (commented):
First part

include/nrf_edgeai_obsv/nrf_edgeai_obsv.h
rrusak, line 1: Please separate these headers from include/nrf_edgeai. This will complicate the nrf_edgeai release process, as the nrf_edgeai library itself doesn't have these headers, and we'll have to constantly copy them whenever I replace the include/nrf_edgeai folder with new release. I suggest separating this into include/nrf_edgeai_obsv and treat it as a separate extension to nrf_edgeai (and not only to this library itself but for all edgeai ecosystem )

include/nrf_edgeai/obsv/nrf_edgeai_obsv.h
```diff
@@ -0,0 +1,126 @@
...
+#ifndef NRF_EDGEAI_OBSV_H
+#define NRF_EDGEAI_OBSV_H
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include <zephyr/kernel.h>
+
+#include <nrf_edgeai/obsv/nrf_edgeai_obsv_metrics.h>
+#include <nrf_edgeai/obsv/nrf_edgeai_obsv_state.h>
```
rrusak: let's keep consistency in naming, if nrf_edgeai_obsv_state.h more about the functionality of nrf_edgeai_obsv_core.c so let's name it nrf_edgeai_obsv_core.h :)

lib/nrf_edgeai_obsv/nrf_edgeai_obsv_core.c
```diff
@@ -0,0 +1,171 @@
...
+		} else if (p_metric->init != NULL) {
+			p_metric->init(NULL);
+		}
+		p_metric = p_metric->p_next;
+	}
+
+	return 0;
+}
+
+int nrf_edgeai_obsv_state_register(nrf_edgeai_obsv_state_t *p_ctx,
```
rrusak: nrf_edgeai_obsv_state_ or nrf_edgeai_obsv_core_? please rename file or functions name (personally I prefer core but the main thing is consistency)

lib/nrf_edgeai_obsv/nrf_edgeai_obsv_core.c
```diff
@@ -0,0 +1,171 @@
...
+#include <nrf_edgeai/obsv/nrf_edgeai_obsv_state.h>
+
+int nrf_edgeai_obsv_state_init(nrf_edgeai_obsv_state_t *p_ctx,
+			       const nrf_edgeai_obsv_model_info_t *p_model)
+{
+	if ((p_ctx == NULL) || (p_model == NULL)) {
+		return -EINVAL;
+	}
+
+	if (p_model->num_classes > CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES) {
```
rrusak: check also if (num_classes == 0 || p_model->num_classes > CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES)

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_probs_distribution.c
```diff
@@ -0,0 +1,100 @@
...
+	float bin_edges[EDGE_NUM];
+	uint32_t counts[CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES][BIN_NUM];
+} state;
+
+/* Stable buffer returned by snapshot(). Copying live counters into this shadow
+ * lets the transport iterate the table without racing against update() for
+ * the full duration of serialization. The race window shrinks to the memcpy
+ * below; see snapshot() doc comment for the residual-race disclaimer.
+ */
+static uint32_t shadow[CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES][BIN_NUM];
```
rrusak: maybe reduce scope of shadow to probs_distribution_snapshot function?

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_probs_distribution.c
```diff
@@ -0,0 +1,100 @@
...
+
+static void probs_distribution_snapshot(nrf_edgeai_obsv_metric_snapshot_t *out)
+{
+	memcpy(shadow, state.counts, sizeof(state.counts));
+
+	out->metric_id = NRF_EDGEAI_OBSV_METRIC_ID_PROBS_DISTRIBUTION;
+	out->version = METRIC_PROBS_DISTRIBUTION_VERSION;
+	out->num_rows = CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES;
+	out->num_cols = BIN_NUM;
+	out->counts = &shadow[0][0];
```
rrusak: I also think this architecture is a bit memory-hungry. If we imagine we have many metrics, and each metric has its own shadow allocated in static memory, then 99% of the time this memory is unused. When encoding, we still process metric after metric, and it would be more efficient if the snapshot used the external memory passed to it. Then we would only need to allocate memory for the largest metric and reuse it when encoding. But for the first iteration, I think this will be fine for us, since there won't be many metrics yet.

lib/nrf_edgeai_obsv_memfault/nrf_edgeai_obsv_memfault.c
```diff
@@ -0,0 +1,209 @@
...
+		return -EINVAL;
+	}
+
+	size_t len = nrf_edgeai_obsv_encode_cbor(&nrf_edgeai_obsv_mflt_staging.ctx->state,
+						 nrf_edgeai_obsv_mflt_staging.buf,
+						 sizeof(nrf_edgeai_obsv_mflt_staging.buf));
+
+	if (len == 0 || len > UINT16_MAX) {
+		k_mutex_unlock(&obsv_mflt_lock);
+		if (ctx != NULL) {
```
rrusak: why we need always to check ctx? Do we need to proceed collect function if nrf_edgeai_obsv_ctx_t *const ctx is null? What we will collect then?

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_probs_distribution.c
mbarchie: Addressed by removing it entirely. The shadow existed to give the transport a stable copy of counters while encoding, but `encode()` already holds `ctx->lock` throughout `for_each_metric()`, which blocks any concurrent `update()`.

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_probs_distribution.c
mbarchie: Good point, and you're right - the shadow was dropped entirely in the updated code. The static per-metric shadow was there to decouple the transport's read path from the live counters, but it might be unnecessary: `encode()` holds `ctx->lock` for its entire duration, which already prevents any concurrent `update()`. Since `snapshot()` is only ever called from `inside for_each_metric()` under that lock, `out->counts` can point directly into the live storage in `priv` with no race and no copy. This also removes the per-metric static RAM overhead you flagged - the only memory cost is the CBOR `tmp` buffer in `collect()`, which is stack-allocated and sized by Kconfig at compile time.

include/nrf_edgeai/obsv/nrf_edgeai_obsv_encode_sizes.h
mbarchie: These are CBOR wire format byte budgets, not C memory layout sizes, so it cannot use sizeof.
I attempted to make macros a little bit more readable, but still probably not perfect.
CBOR is variable-length: e.g. a uint32_t with value 0 encodes as 1 byte, not 5. The macros compute worst-case encoded byte counts (e.g. `_NRF_EDGEAI_OBSV_ENCODE_UINT32_WORST = 1 + 4`), which don't correspond to any struct in memory.

Regarding previous comment - it doesn't need to be public, I moved this header to `lib/nrf_edgeai_obsv_memfault/`, so it acts as private header.

mbarchie:
I updated the PR with implementation allowing to gather metrics for more than 1 model.

mbarchie:
FYI - I added another commit (https://github.com/nrfconnect/sdk-edge-ai/pull/83/changes/01fd45e82f385f6d38ec5cb2ddb14c6dbda96d82) that replaces maybe a little bit over-complicated static storage macros with `_create` metrics functions. I think it simplifies maintanence/usage at low cost. It should be fairly easy to compare with previous implementation as I put obsv_cor implementation in separate commit.

Further simplification could be to assume worst case size scenario for each metric type - e.g. all metrics can handle `CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES`. Then metrics become e.g.

```
typedef struct {
    _nrf_obsv_tm_hdr_t hdr;
    uint32_t matrix[CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES][CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES];
} _nrf_obsv_tm_storage_t;
```

and you easily iterate over compile-time defined data fields of metrics. The trade-off is obviously memory consumption.

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
```diff
@@ -0,0 +1,238 @@
...
+	NRF_EDGEAI_OBSV_METRIC_ID_TRANSITION_MATRIX = 2,
+	NRF_EDGEAI_OBSV_METRIC_ID_PROBS_DISTRIBUTION = 3,
+};
+
+/**
+ * @brief Header (dimension fields) for the probability distribution metric storage.
+ *
+ * Shared between the storage macro and the metric implementation so the layout
+ * is defined in exactly one place. The explicit _pad[2] aligns the float array
+ * that immediately follows to a 4-byte boundary; sizeof(_nrf_obsv_pd_hdr_t) == 4.
```
Szynkaa: build assert on the sizeof

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * Not intended for direct use outside of the metric implementation.
+ */
+typedef struct {
+	uint8_t num_classes;
```
Szynkaa: in the lib `num_classes` in hold in uin16_t (nrf_edgeai_output_types.h) - is this mismatch intentional?

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
```diff
@@ -0,0 +1,238 @@
...
+ * Pass to nrf_edgeai_obsv_core_register() as the p_cfg argument.
+ * NULL p_cfg gives uniform bins over [0, 1].
+ */
+typedef struct {
+	/**
+	 * Array of (CONFIG_NRF_EDGEAI_OBSV_PROBS_DISTRIBUTION_BIN_NUM + 1) edge values
+	 * in ascending order over [0, 1], e.g. {0.0, 0.25, 0.5, 0.75, 1.0}
+	 * for 4 uniform bins.
+	 */
+	float bin_edges[CONFIG_NRF_EDGEAI_OBSV_PROBS_DISTRIBUTION_BIN_NUM + 1];
```
Szynkaa: are we expecting that someone will need bins not covering the full probability range? ergo, are the first (0.0) and last (1.0) edges needed?

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_probs_distribution.c
```diff
@@ -0,0 +1,122 @@
...
+
+	compute_uniform_edges(edges, hdr->bin_num);
+}
+
+static void pd_update(const float *p_probs, uint8_t n, void *priv)
+{
+	const _nrf_obsv_pd_hdr_t *hdr = priv;
+	const float *edges = pd_edges(hdr);
+	uint32_t *counts = pd_counts(hdr);
+	uint8_t classes = (n < hdr->num_classes) ? n : hdr->num_classes;
```
Szynkaa: min

Szynkaa: or rather assert as the metric is implemented for specific number of classes?

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_transition_matrix.c
```diff
@@ -0,0 +1,105 @@
...
+#include <string.h>
+
+#include <nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h>
+
+#define METRIC_TRANSITION_MATRIX_VERSION 1
+
+/* Sentinel stored in prev when no inference has been seen yet. Must be outside
+ * the valid class index range [0, num_classes), which always fits in uint8_t.
+ */
+#define NO_PREV_CLASS 0xFF
```
Szynkaa: mismatch with `num_classes` in nRF Edge AI Lib

Szynkaa: 0xFF fits in uint8_t

lib/nrf_edgeai_obsv/Kconfig
```diff
@@ -0,0 +1,63 @@
...
+	default 0
+	help
+	  Additional bytes added to NRF_EDGEAI_OBSV_ENCODE_MAX_SIZE to
+	  accommodate custom metrics registered alongside the built-in ones.
+	  Set to at least the sum of NRF_EDGEAI_OBSV_ENCODE_METRIC_SIZE(n_rows,
+	  n_cols) evaluated for each custom metric. The Memfault staging buffer
+	  encode buffer are sized from NRF_EDGEAI_OBSV_ENCODE_MAX_SIZE,
+	  so they grow automatically.
+
+config NRF_EDGEAI_OBSV_MAX_CLASSES
```
Szynkaa, line 35: range 1 65535

lib/nrf_edgeai_obsv/nrf_edgeai_obsv_core.c
```diff
@@ -0,0 +1,169 @@
...
+
+	p_metric->p_next = NULL;
+
+	if (p_tail == NULL) {
+		p_ctx->p_metrics_list = p_metric;
+	} else {
+		p_tail->p_next = p_metric;
+	}
+
+	p_ctx->num_metrics++;
```
Szynkaa, line 88: potential overflow

lib/nrf_edgeai_obsv_memfault/Kconfig
```diff
@@ -0,0 +1,64 @@
...
+	  (BT_MDS, MEMFAULT_HTTP_PERIODIC_UPLOAD) and the application
+	  does not want to wire a timer itself. The first collect runs
+	  one interval after initialization.
+
+	  When enabled (and whenever collect can run concurrently with inference),
+	  bind the transport with nrf_edgeai_obsv_memfault_init and drive
+	  inference through nrf_edgeai_obsv_update (see nrf_edgeai_obsv.h)
+	  so encode inside collect does not race snapshot collection.
+
+config NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT_INTERVAL_SEC
```
Szynkaa, line 51: Does this conflict with memfault limit of 1 CDR per day?

lib/nrf_edgeai_obsv_memfault/nrf_edgeai_obsv_memfault.c
```diff
@@ -0,0 +1,264 @@
...
+	for (uint8_t i = 0; i < nrf_edgeai_obsv_mflt_staging.num_ctxs; i++) {
+		if (nrf_edgeai_obsv_mflt_staging.ctxs[i] == ctx) {
+			k_mutex_unlock(&obsv_mflt_lock);
+			return -EALREADY;
+		}
+	}
+
+	if (nrf_edgeai_obsv_mflt_staging.num_ctxs >= CONFIG_NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS) {
+		k_mutex_unlock(&obsv_mflt_lock);
+		LOG_ERR("init: max contexts (%d) reached",
```
Szynkaa, line 154: logging subsys has prepending function name as a feature

lib/nrf_edgeai_obsv_memfault/nrf_edgeai_obsv_memfault.c
```diff
@@ -0,0 +1,264 @@
...
+
+	/*
+	 * Build the CDR payload into a temporary buffer as a CBOR array:
+	 *
+	 *   [obsv-payload-ctx0, obsv-payload-ctx1, ...]
+	 *
+	 * The outer array header is written manually; nrf_edgeai_obsv_encode()
+	 * acquires ctx->lock internally so obsv_mflt_lock must not be held here.
+	 */
+	uint8_t tmp[NRF_EDGEAI_OBSV_ENCODE_CDR_STAGING_BUFSZ];
```
Szynkaa: this force enlarging stack, potentially a lot - at least inform about it, but I'm not sure where

Szynkaa (commented):
Part 2

tests/obsv_memfault/testcase.yaml
```diff
@@ -0,0 +1,20 @@
...
+  tags: edgeai obsv memfault
+  platform_allow:
+    - qemu_cortex_m3
+  integration_platforms:
+    - qemu_cortex_m3
+
+tests:
+  edgeai.obsv_memfault: {}
+
+  edgeai.obsv_memfault.auto_collect:
```
Szynkaa, line 18: how is this config tested?

tests/obsv_memfault/src/main.c
```diff
@@ -0,0 +1,328 @@
...
+
+/* ---------- Overwrite behavior ---------- */
+
+/*
+ * collect() must overwrite the staging buffer even when the previous CDR has
+ * not yet been drained. The packetizer's read_data() bounds-checks against
+ * staging.len (updated atomically under obsv_mflt_lock), so stale reads are
+ * caught gracefully. Fresh data is always preferred over stale.
+ */
+ZTEST(obsv_memfault, test_collect_overwrites_when_not_drained)
```
Szynkaa, line 195: This test does not verify that collect actually overwrote the buffer

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
mbarchie: I agree, they're not used anywhere right now and can be added if ever needed

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
mbarchie: I added standard C static asserts

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
mbarchie: Aligned to `uint16_t`, to match Edge AI Lib's capabilities. In `Kconfig`, I added a note about memory consumption as max num of classes grows.

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_probs_distribution.c
mbarchie: I'll put an assert as it's rather programmer's error than run-time condition.

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_transition_matrix.c
mbarchie: With number of classes in range [1, 2, ..., 255], you get indices [0, 1, ..., 254], so 255 is not a valid class ID.

I extended it to `uint16_t` to match with Edge AI Lib and changed this macro to `0xFFFF`.

lib/nrf_edgeai_obsv_memfault/Kconfig
mbarchie: I should probably change it 1 per day as default value not to make confusion

However, the limit can be made higher selectively by contacting support or using device in developer mode. I updated also the Kconfig note.

lib/nrf_edgeai_obsv_memfault/nrf_edgeai_obsv_memfault.c
mbarchie: I dropped per-function name prefixes in favor of a uniform `CDR:` context prefix on all `LOG_DBG` calls. `CONFIG_LOG_FUNC_NAME_PREFIX_DBG` is a build-time opt-in outside our library's control, so we don't rely on it.

lib/nrf_edgeai_obsv_memfault/nrf_edgeai_obsv_memfault.c
mbarchie: I added a note both in `NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT` Kconfig option and in doxygen for `nrf_edgeai_obsv_memfault_collect`

tests/obsv_memfault/testcase.yaml
mbarchie: The `auto_collect` scenario shares the tests with base test. It verified only that the code compiles and the CDR API works correctly with `AUTO_COLLECT=y`. The work item itself doesn't fire during the test run.

I added a test that calls `auto_collect_work_handler()` directly (exposed via `STATIC_EXCEPT_TEST`). The `auto_collect` scenario now has one extra test case that verifies the handler actually stages a CDR payload.

tests/obsv/src/suite_user_storage.c
mbarchie, line 1: I missed adding this test suite in previous commits

lib/nrf_edgeai_obsv_memfault/Kconfig
```diff
@@ -40,17 +40,30 @@ config NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT
...
+	  nrf_edgeai_obsv_encode_sizes.h for the exact value at your Kconfig
+	  dimensions). When calling collect() manually, ensure your thread's
+	  stack is large enough.
+
 config NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT_INTERVAL_SEC
 	int "Auto-collect interval (seconds)"
 	depends on NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT
-	default 3600
+	default 86400
 	range 60 86400
```
Szynkaa: why we set 1 day upper limit for interval?

Szynkaa (commented):
I assume the "fix" commit will be squashed/fixuped. Address compliance, rather easy fix.

lib/nrf_edgeai_obsv_memfault/Kconfig
mbarchie: I changed it make it work with default memfault limitation for CDR.

mbarchie:
> I assume the "fix" commit will be squashed/fixuped. Address compliance, rather easy fix.

I'll make a casting "workaround" for compliance check.

Regarding the commit history - I'll rewrite it properly once I get approvals for the content of the PR.

lib/nrf_edgeai_obsv_memfault/Kconfig
mbarchie: Clarified off-line - I will change upper limit to a higher value (e.g. 30 days will fit into uin32_t)

mbarchie:
Pushing a batch of follow-up changes addressing reviewer feedback and improving test architecture. Summary of what changed:

---

### Encoding moved to `obsv` core

`nrf_edgeai_obsv_encode_list()` is now part of the `nrf_edgeai_obsv` library itself (not the Memfault transport).
The function encodes N contexts as a single `[+ obsv-payload]` CBOR array - it can be used with any transport, not just Memfault.

The Memfault transport now simply calls `encode_list()` to fill the CDR staging buffer; no CBOR logic lives in the transport layer anymore.

Sizing macros in `nrf_edgeai_obsv_encode_sizes.h` were renamed from `CDR`-prefixed to `LIST`-prefixed to reflect transport-agnostic scope (e.g. `NRF_EDGEAI_OBSV_ENCODE_LIST_BUF_SIZE(n)`).

The CDDL schema (`obsv.cddl`) was moved into `lib/nrf_edgeai_obsv/` alongside the encoder.

---

### Test isolation for `obsv_memfault`

`tests/obsv_memfault` now tests only the Memfault CDR transport layer. Two mocks replace all external dependencies:

| Mock | What it replaces |
|---|---|
| `memfault_mock.c` | `memfault_cdr_register_source()` (the only Memfault SDK function called at init) |
| `obsv_mock.c` | `nrf_edgeai_obsv_encode_list()` — mock tracking call count, returns `n × 16` bytes stamped with it |

Consequences:
- `CONFIG_MEMFAULT=y` is no longer set in `prj.conf`; the Memfault include path is injected by `CMakeLists.txt` directly. No Memfault SDK sources compile - avoids arch-specific panics/coredump code that broke `native_sim/native/64`.
- `CONFIG_NRF_EDGEAI_OBSV_ENCODE=y` / `CONFIG_ZCBOR=y` are no longer required in the transport test; the encoder is fully stubbed.
- Tests now run on **`native_sim/native/64`** (integration platform) and `qemu_cortex_m3`.

---

### CDDL conformance test moved to `tests/obsv`

`suite_encoder.c` (new) in `tests/obsv` compiles `obsv.cddl` to a zcbor decoder at configure time and validates live encoder output against the schema. This belongs next to the encoder, not the transport.

---

### Bug fix: metric type guards in `nrf_edgeai_obsv_metrics.h`

Probability distribution and transition matrix types, storage macros, and `_create()` declarations were unconditionally defined but referenced Kconfig symbols only available when the respective metric was enabled. Wrapped each family in its controlling Kconfig guard (`CONFIG_NRF_EDGEAI_OBSV_METRIC_PROBS_DISTRIBUTION` / `CONFIG_NRF_EDGEAI_OBSV_METRIC_TRANSITION_MATRIX`).

---

### Auto-collect interval upper limit raised to 30 days

`NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT_INTERVAL_SEC` range changed from `60 86400` (1 day) to `60 2592000` (30 days). 30 days fits safely within `uint32_t` after `s * 1000U`.

---

include/nrf_edgeai_obsv/nrf_edgeai_obsv.h
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * @param ctx  Initialized context.
+ * @param cb   Callback invoked once per metric snapshot. Return false to stop early.
+ * @param user Opaque pointer forwarded to @p cb.
+ * @return 0 on success, -EINVAL if @p ctx or @p cb is NULL.
+ */
+int nrf_edgeai_obsv_for_each_metric(nrf_edgeai_obsv_ctx_t *ctx, nrf_edgeai_obsv_metric_cb_t cb,
+				    void *user);
+
+#if defined(CONFIG_NRF_EDGEAI_OBSV_ENCODE)
```
Szynkaa: Zephyr does not limit their API declarations if some module is disabled, maybe remove this if guards?

lib/nrf_edgeai_obsv_memfault/obsv.cddl
Szynkaa, line 1: to be removed?

Szynkaa (approved):

mbarchie:
the last push was a rebase

rrusak (approved):

mbarchie:
@annwoj Can you review text in Kconfigs?

annwoj (approved):

PR #97, opened 2026-05-08
treewide: opt out of deprecated Partition Manager

grochu:
Partition Manager has been deprecated in NCS v3.3 in favor of devicetree-based flash partitioning. It is still enabled by default for nRF52/53/91 and the nRF54L family, so every project here was implicitly using it.

Add a project-level Kconfig.sysbuild that overrides the default with "config PARTITION_MANAGER default n" for all applications, samples and tests. This covers all supported board targets in one place, matching the pattern used by upstream NCS samples.

No pm_static.yml or pm.yml files exist in the repository, so partition layouts now come straight from the board devicetree.

### DIFF
--- a/applications/gesture_recognition/configuration/nrf54l15tag_nrf54l15_cpuapp/images/mcuboot/nrf54l15tag_nrf54l15_cpuapp.overlay
+++ b/applications/gesture_recognition/configuration/nrf54l15tag_nrf54l15_cpuapp/images/mcuboot/nrf54l15tag_nrf54l15_cpuapp.overlay
@@ -4,15 +4,9 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 
-/ {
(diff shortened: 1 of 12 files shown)


mbarchie (approved):

mbarchie (approved):

applications/gesture_recognition/Kconfig.sysbuild
```diff
@@ -4,6 +4,9 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
+config PARTITION_MANAGER
+	default n if !BOARD_THINGY53
```
pioso-dev, line 8: This exception made for thingy53 is considered intentional and acceptable despite task title

pioso-dev (approved):
In gesture recognition application the exception made for thingy53 is considered intentional even if against task title
Approved

PR #108, opened 2026-06-08
samples: data_forwarder: Add sensor streaming sample

Szynkaa:
Introduce a sample that streams sensor data to a host application over
BLE NUS or UART. Samples are framed with CBOR, with extra COBS wrapping
for transport without message boundries; the message schema is defined
in CDDL and encoded with zcbor so hosts can decode traffic without
device-specific knowledge.

Session metadata (sampling rate, channel layout, sensor type, and device
name) is sent at session start and re-sent periodically for late joiners.
BMI270, ADXL367, and BME688 backends are selectable at build time, with
optional BME688 upsampling when a motion sensor is the primary source.

Jira: NCSDK-38292


samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+#include <string.h>
+
+#include <zephyr/data/cobs.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/net_buf.h>
+#include <zephyr/random/random.h>
+#include <zephyr/sys/atomic.h>
+#include <zephyr/sys/crc.h>
+#include <zephyr/sys/printk.h>
```
mbarchie: `printk` not needed

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+			continue;
+		}
+
+		strncpy(channel_names[i], cfg->channel_names[i], PROTO_CHANNEL_NAME_MAX_LEN);
+		channel_names[i][PROTO_CHANNEL_NAME_MAX_LEN] = '\0';
+	}
+
+	session_cfg = *cfg;
+	strncpy(session_device_name, cfg->device_name, PROTO_DEVICE_NAME_MAX_LEN);
+	session_device_name[PROTO_DEVICE_NAME_MAX_LEN] = '\0';
```
mbarchie: `snprintk` can be used instead as an one-liner

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	for (size_t i = 0; i < cfg->channels; i++) {
+		if (cfg->channel_names[i] == NULL) {
+			continue;
+		}
+
+		strncpy(channel_names[i], cfg->channel_names[i], PROTO_CHANNEL_NAME_MAX_LEN);
+		channel_names[i][PROTO_CHANNEL_NAME_MAX_LEN] = '\0';
+	}
```
mbarchie, line 468: Is `channel_names` needed at all? Whole `cfg` is copied to `session_cfg` anyway and it seems that one is used in the rest of the code.

samples/data_forwarder/prj.conf
```diff
@@ -0,0 +1,49 @@
...
+
+# Select int32 for use with Edge AI Lab
+CONFIG_DATA_FWD_PROTO_INT32_VALUES=y
+
+# Data forwarder dependencies
+CONFIG_ZCBOR=y
+CONFIG_COBS=y
+CONFIG_CRC=y
+
+# Sensor dependecies
```
mbarchie:
```suggestion
# Sensor dependencies
```

samples/data_forwarder/prj.conf
```diff
@@ -0,0 +1,49 @@
...
+CONFIG_BT_PERIPHERAL=y
+CONFIG_BT_DEVICE_NAME="nRF DataFwd"
+CONFIG_BT_MAX_CONN=1
+CONFIG_BT_NUS=y
+CONFIG_BT_GATT_CLIENT=y
+CONFIG_BT_L2CAP_TX_MTU=247
+CONFIG_BT_BUF_ACL_RX_SIZE=251
+
+# C library and runtime support
+CONFIG_FPU=y
```
mbarchie: Do we need `CONFIG_FPU` if `CONFIG_DATA_FWD_PROTO_INT32_VALUES` is enabled?

samples/data_forwarder/prj.conf
```diff
@@ -0,0 +1,49 @@
...
+CONFIG_CRC=y
+
+# Sensor dependecies
+CONFIG_SENSOR=y
+CONFIG_BMI270=y
+CONFIG_BME680=y
+
+# Logging and assert
+CONFIG_LOG=y
+CONFIG_ASSERT=y
```
mbarchie, line 32: These configs are duplicated in lines 44-49

samples/data_forwarder/src/protocol/protocol_types.h
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+#ifndef DATA_FORWARDER_PROTOCOL_TYPES_H_
+#define DATA_FORWARDER_PROTOCOL_TYPES_H_
+
+#if CONFIG_DATA_FWD_PROTO_INT32_VALUES
+#define PROTO_VALUE_T int32_t
+#else
+#define PROTO_VALUE_T float
+#endif
```
mbarchie, line 22: IMO using `typedef` would be more idiomatic

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+
+	uint32_t put = ring_buf_put(&proto_rb, (uint8_t *)&sample, sizeof(sample));
+
+	if (put != sizeof(sample)) {
+		ring_buf_reset(&proto_rb);
+		k_mutex_unlock(&proto_rb_lock);
+
+		atomic_inc(&producer_drops);
+		return -EFAULT;
+	}
```
mbarchie, line 543: This branch discards all the samples in the ring buffer. Thus, incrementing `producer_drops` by 1 is not fully correct.

This execution path is rather unlikely to happen (rather some ring buffer/memory corruption thing), but I guess adding `queued_samples + 1` would be more correct operation to do.

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+	net_buf_push_le16(buf, buf->len);
+
+	if (IS_ENABLED(CONFIG_DATA_FWD_PROTO_CRC)) {
+		const uint16_t crc = crc16_ccitt(0xFFFF, buf->data + PROTO_LENGTH_SIZE,
+						 buf->len - PROTO_LENGTH_SIZE);
+
+		net_buf_add_le16(buf, crc);
+	}
+
+	if (!transport_cfg.has_message_boundaries) {
```
mbarchie: The implementation is correct, but the `tx_buf = buf` alias makes ownership harder to follow.

Consider splitting the COBS and direct-send paths so each branch owns its own `net_buf_send`/`net_buf_unref`.

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+	return err;
+}
+
+static void proto_data_work_handler(struct k_work *work)
+{
+	ARG_UNUSED(work);
+
+	int err;
+
+	struct proto_sample sample;
```
mbarchie: Minor: extra line between definitions, unlike the other functions

```suggestion
	int err;
	struct proto_sample sample;
```

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+		return -EINVAL;
+	}
+
+	for (size_t i = 0; i < sample->count; i++) {
+		if (IS_ENABLED(CONFIG_DATA_FWD_PROTO_INT32_VALUES)) {
+			n = snprintk(line + off, remain, "%s%d", i == 0 ? "" : ",",
+				     (int32_t)sample->values[i]);
+		} else {
+			n = snprintk(line + off, remain, "%s%f", i == 0 ? "" : ",",
+				     (double)sample->values[i]);
```
mbarchie, line 194: Extracting CSV line encoding into separate helper function would make this part more self-explanatory. It would also separate encoding vs transport and align with CBOR-related encoding functions for non-ASCII mode.

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+		if ((n < 0) || (n >= (remain))) {
+			return -EMSGSIZE;
+		}
+
+		off += n;
+		remain -= n;
+	}
+
+	n = snprintk(line + off, remain, "\r\n");
+	if ((n < 0) || (n >= (remain))) {
```
mbarchie: minor:
```suggestion
	if ((n < 0) || (n >= remain)) {
```

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+		k_work_submit_to_queue(&proto_work_q, &proto_data_work);
+	}
+
+	if (IS_ENABLED(CONFIG_DATA_FWD_PROTO_ASCII_MODE)) {
+		err = proto_send_ascii(&sample);
+		if (err) {
+			LOG_WRN_RATELIMIT("Samples send failed: %d", err);
+		}
+
+		return;
```
mbarchie, line 347: Error here won't increment `producer_drops`. For consistency with non-ASCII mode, it might worth adding counter here too.

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+
+		if (err) {
+			LOG_ERR("COBS encode failed (err %d)", err);
+			net_buf_unref(tx_buf);
+
+			return err;
+		}
+	}
+
+	err = transport_cfg.send(tx_buf->data, tx_buf->len, transport_cfg.ctx);
```
mbarchie: Should we increment `producer_drops` here too? Or is it a way of distinguishing between types of lost samples? This comment applies also to failures in ASCII mode.

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,456 @@
...
+		k_mutex_unlock(&proto_rb_lock);
+		return;
+	}
+
+	if (got != sizeof(sample)) {
+		ring_buf_reset(&proto_rb);
+		k_mutex_unlock(&proto_rb_lock);
+
+		return;
+	}
```
mbarchie: Dropping ring buffer content should be logged

samples/data_forwarder/src/sensor/adxl367.c
```diff
@@ -0,0 +1,128 @@
...
+		k_work_submit(&bme688_work_wrapper.work);
+	}
+#endif
+
+	err = sensor_channel_get(dev_adxl, SENSOR_CHAN_ACCEL_XYZ, accel);
+	if (err) {
+		return err;
+	}
+
+	for (size_t i = 0; i < 3; i++) {
```
mbarchie:
```suggestion
	for (size_t i = 0; i < ADXL367_CHANNEL_COUNT; i++) {
```

samples/data_forwarder/src/sensor/adxl367.c
```diff
@@ -0,0 +1,128 @@
...
+#endif
+
+	return 0;
+}
+
+int data_fwd_sensor_fetch(PROTO_VALUE_T *values, const size_t values_size, size_t *count)
+{
+	int err;
+
+	struct sensor_value accel[3] = {0};
```
mbarchie:
```suggestion
	struct sensor_value accel[ADXL367_CHANNEL_COUNT] = {0};
```

samples/data_forwarder/src/protocol/protocol.h
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @retval 0 Success.
+ * @retval -EINVAL @p transport or @p transport->send is @c NULL.
+ */
+int proto_init(const struct proto_transport *transport);
+
+/**
+ * @brief Start a new streaming session.
+ *
+ * Copies @p cfg, assigns a random session ID, and begins sending session-info messages.
```
mbarchie, line 74: It actually does a shallow copy of `cfg`, may be worth mentioning or changing

samples/data_forwarder/src/sensor/data_fwd_sensor.h
```diff
@@ -0,0 +1,92 @@
...
+uint8_t data_fwd_sensor_type_id(void);
+
+/**
+ * @brief Get the nominal sampling frequency in Hz.
+ *
+ * @return Sampling rate in Hz for the selected sensor.
+ */
+uint16_t data_fwd_sensor_frequency(void);
+
+static inline PROTO_VALUE_T data_fwd_sensor_value_to_proto_value(const struct sensor_value *val)
```
mbarchie: Missing doc comment

samples/data_forwarder/src/sensor/data_fwd_sensor.h
```diff
@@ -0,0 +1,92 @@
...
+ */
+uint8_t data_fwd_sensor_channel_count(void);
+
+/**
+ * @brief Get short names for each data channel.
+ *
+ * The returned array has @ref data_fwd_sensor_channel_count() entries and remains valid
+ * for the lifetime of the application.
+ *
+ * @return Pointer to a null-terminated array of channel name strings.
```
mbarchie: It sounds like a single string. Shouldn't it be "an array of null-terminated channel name strings"?

samples/data_forwarder/src/protocol/protocol.h
```diff
@@ -0,0 +1,114 @@
...
+ */
+struct proto_session_config {
+	/** Sampling rate in Hz. */
+	uint16_t rate_hz;
+	/** Number of data channels in each sample. */
+	uint8_t channels;
+	/** Application-defined sensor type identifier. */
+	uint8_t sensor_type;
+	/** Names of the @c channels data channels (one string per channel). */
+	const char *const *channel_names;
```
mbarchie: Max length should be documented here

samples/data_forwarder/src/protocol/protocol.h
```diff
@@ -0,0 +1,114 @@
...
+ * If the ring buffer is full, the oldest sample is dropped and the drop
+ * counter in session-info messages is incremented.
+ *
+ * @param values  Channel values (length @p count).
+ * @param count   Number of values; must not exceed @c CONFIG_DATA_FWD_PROTO_MAX_CHANNELS.
+ *
+ * @retval 0 Success.
+ * @retval -EACCES No session is active.
+ * @retval -EINVAL @p values is @c NULL or @p count is zero or too large.
+ * @retval -EFAULT Internal buffer was corrupted. @p sample dropped.
```
mbarchie:
```suggestion
 * @retval -EFAULT Internal buffer was corrupted. @p values are dropped.
```

samples/data_forwarder/Kconfig
```diff
@@ -0,0 +1,113 @@
...
+	depends on ADXL367
+
+config DATA_FWD_SENSOR_BME688
+	bool "BME688 (environmental)"
+	depends on SENSOR
+	depends on BME680
+
+endchoice
+
+config DATA_FWD_EXTRA_SENSOR_BME688
```
mbarchie, line 98: This item lacks a help message to explain what it really does

samples/data_forwarder/src/sensor/bmi270.c
```diff
@@ -0,0 +1,197 @@
...
+
+	for (size_t i = 0; i < 3; i++) {
+		values[i] = data_fwd_sensor_value_to_proto_value(&accel[i]);
+		values[i + 3] = data_fwd_sensor_value_to_proto_value(&gyro[i]);
+	}
+
+	*count = BMI270_CHANNEL_COUNT;
+
+#if IS_ENABLED(CONFIG_DATA_FWD_EXTRA_SENSOR_BME688)
+	err = data_fwd_bme688_append(dev_bme, values + *count, values_size - *count, count);
```
mbarchie, line 173: In theory values from BME688 can still be stale when appending (async readout may not be completed). Not a big deal given the data rate, but may be worth mentioning this data race.

samples/data_forwarder/src/sensor/bmi270.c
```diff
@@ -0,0 +1,197 @@
...
+	/* Set sampling frequency last as this also sets the appropriate power mode. If already
+	 * sampling, change sampling frequency to 0.0Hz before changing other attributes
+	 */
+	ok &= bmi_set_attr(SENSOR_CHAN_GYRO_XYZ, SENSOR_ATTR_SAMPLING_FREQUENCY, &sampling_freq);
+
+	if (!ok) {
+		return -ENOTSUP;
+	}
+
+	const uint32_t period_ns = Z_HZ_ns / MAX(1, FREQUENCY_HZ);
```
mbarchie: If I understand this correctly, `MAX(1, FREQUENCY_HZ)` is to avoid division by zero? If so, build assert would be more intuitive way of keeping non-zero frequency.

samples/data_forwarder/src/sensor/bmi270.c
```diff
@@ -0,0 +1,197 @@
...
+#include <zephyr/logging/log.h>
+
+#include "data_fwd_sensor.h"
+#include "bme688_common.h"
+
+LOG_MODULE_REGISTER(sensor);
+
+#define BMI270_NODE	     DT_NODELABEL(bmi270)
+#define BMI270_CHANNEL_COUNT 6
+#define SENSOR_TYPE_VALUE    1
```
mbarchie, line 17: If sensor IDs are supposed not to overlap, I'd keep all of them in one place (e.g. `sensor_types.h`)

samples/data_forwarder/Kconfig
```diff
@@ -0,0 +1,113 @@
...
+config DATA_FWD_TRANSPORT_BLE_NUS
+	bool "BLE NUS"
+	depends on BT
+	depends on BT_PERIPHERAL
+	depends on BT_NUS
+
+config DATA_FWD_TRANSPORT_UART
+	bool "UART"
+	depends on SERIAL
+	depends on UART_ASYNC_API
```
mbarchie, line 74: Is UART config testable out-of-the-box or rather just prepared when working on sample on different DK than l15tag?

samples/data_forwarder/src/protocol/protocol.c
Szynkaa: `producer_drop` was intended to count samples dropped due to errors in protocol, not in transport. I will fix all usage of it

samples/data_forwarder/src/protocol/protocol.c
Szynkaa: as above

samples/data_forwarder/src/protocol/protocol.h
Szynkaa: Which part is copied shallowly? Scalar elements are copied in `session_cfg = *cfg;`, device and channel names are copied to internal string buffers

samples/data_forwarder/src/sensor/bmi270.c
Szynkaa: They are not overlapping, but IMO creating single file would create sense of direct connection between sensor and ID. And this connection is not true as there is ability to extend base sensor with values from BME688. This IDs are to fill the protocol field for sensor type, and this field is purely for use by someone using the protocol (aka custom data).

samples/data_forwarder/src/protocol/protocol.h
Szynkaa: I agree that there is a bug in the use of the copied elements

samples/data_forwarder/src/protocol/protocol.c
Szynkaa: Used for deep copy, usage fixed

samples/data_forwarder/Kconfig
Szynkaa: It is not testable out-of-the-box as this board does not have any easy to use UART interfaces. Documentation of sample will note. Building sample for tag will fail due to missing chosen node

samples/data_forwarder/src/sensor/bmi270.c
mbarchie: I agree with your reasoning, fine to leave as-is

samples/data_forwarder/Kconfig
mbarchie: I thought so - good that README will cover it

samples/data_forwarder/src/protocol/protocol.c
```diff
@@ -0,0 +1,550 @@
...
+		return -EINVAL;
+	}
+
+	if (cfg->channels > CONFIG_DATA_FWD_PROTO_MAX_CHANNELS) {
+		LOG_ERR("Increase CONFIG_DATA_FWD_PROTO_MAX_CHANNELS to accommodate %u channels",
+			cfg->channels);
+		return -ENOMEM;
+	}
+
+	/* new version */
```
mbarchie: to be removed?

mbarchie (approved):

samples/data_forwarder/sample.yaml
grochu, line 1: Can we test also the UART transport on the twister level: `DATA_FWD_TRANSPORT_UART`?

grochu (commented):
Looks good, just a comment on the test scope.

grochu (approved):

samples/data_forwarder/sample.yaml
Szynkaa: no, because the supported target does not have predefined UART to use, we could include it in the overlay, but I think this would be misleading

PR #112, opened 2026-06-19
tools: data_forwarder_host: Add sensor data forwarder host GUI

pioso-dev:
## Description
This PR adds `data_forwarder_host`, a new cross-platform desktop GUI tool that receives, visualises, records and exports sensor data forwarded from an nRF device over **UART** or **BLE NUS** (Nordic UART Service). It is the host-side counterpart to the `data_forwarder` device sample and decodes the device's COBS + CBOR v1 frame format. Multiple capture sessions can run side by side, each in its own tab.

## Highlights
- **Sources**: UART (pyserial) and BLE NUS (bleak), with a New Session dialog that scans for and connects to advertising BLE devices.
- **Out-of-process acquisition**: byte reading + COBS/CBOR decode run in a `spawn` child process so the GUI stays responsive regardless of backend load. The GUI process drains decoded frames over an IPC queue at its own cadence with bounded, consumer-paced back-pressure that never drops control/metadata frames. Set `DFH_SINGLE_PROCESS=1` to fall back to the in-GUI path.
- **Visualisation**: combined overlay plot and collapsible per-channel charts (QtCharts), render-time min/max decimation, rolling time window, recording start/stop markers, and live ASCII / decoded-frame log consoles.
- **Recording**: RAM-buffered capture dumped to CSV on stop (written incrementally off the event loop), each paired with a `{stem}.txt` metadata sidecar (host, transport, timing, device session info, channels, errors). A persistent banner shows the last saved file with open-file / open-folder shortcuts.
- **Diagnostics**: live bandwidth/throughput readout, host CPU/RAM metrics, per-category error & loss analysis with transport-loss confirmation windows, First Failure Data Capture, and an animated pipeline-flow view of stage utilisation/back-pressure.
- **Architecture**: strictly layered (`platform` -> `source` -> `protocol` -> `pipeline` -> `core` -> `session` -> `gui`). The acquisition/transport layers (`platform`, `source`, `protocol`, `pipeline`) are Qt-free and run in the headless child process; the data-model, session and GUI layers run in the GUI process and use Qt.

## How to try it (recommended)
The easiest, dependency-free way to run and test the app — and the default way to try this PR — is to build the standalone single-file binary:

```bash
tools/data_forwarder_host$ ./scripts/build_linux_binary.sh
```

This provisions an isolated venv and runs PyInstaller, emitting the single file:

```
dist/data-forwarder-host
```

Run it directly (no Python or dependencies required on the machine):

```bash
tools/data_forwarder_host$ ./dist/data-forwarder-host
```

## How to run (from source)
```bash
cd tools/data_forwarder_host
pip install -r requirements.txt
data-forwarder-host        # or: python -m data_forwarder_host / python main.py
```

Requires Python 3.12+ on Linux, Windows or macOS. PySide6 (Qt 6 with QtCharts) is a base dependency and is always installed. On Linux, UART sources need serial-port access (usually `dialout` group membership); on Ubuntu/Debian the Qt xcb platform plugin also needs `sudo apt install libxcb-cursor0`.

The matching device-side firmware is the `data_forwarder` sample (`edge-ai/samples/data_forwarder`); without it running there is no data to receive.

## Verification
Manually verified UART and BLE NUS capture, multi-session tabs, recording → CSV + metadata sidecar, and GUI responsiveness under flood (out-of-process path) using the `scripts/repro_freeze.py` harness. Validated on Ubuntu; macOS and Windows are expected to run via bleak/CoreBluetooth/WinRT but are not yet validated.

## Notes
- All files are new additions; no existing code is modified.
- Every file carries the SPDX header `LicenseRef-Nordic-5-Clause`.

## Jira
NCSDK-39472

### DIFF
--- a/tools/data_forwarder_host/README.md
+++ b/tools/data_forwarder_host/README.md
@@ -71,6 +71,34 @@ By default, source acquisition (byte reading + COBS/CBOR decode) runs in a
 separate child process so the GUI stays responsive regardless of backend load.
 Set `DFH_SINGLE_PROCESS=1` to fall back to the legacy in-GUI acquisition path.
 
+## Standalone single-file binary (no install on the target)
+
+To run the application on a machine that has **no Python and no dependencies
+installed**, build a single self-contained executable with PyInstaller:
+
+```bash
+./scripts/build_linux_binary.sh
+```
+
+This produces one file, `dist/data-forwarder-host`, that bundles the Python
+interpreter, Qt6 and every dependency. Copy that single file to any reasonably
+recent x86_64 Ubuntu machine and run it directly — nothing has to be installed
+there:
+
+```bash
+./data-forwarder-host
+```
+
+Notes:
+
+- Build on the **oldest** Ubuntu you intend to support: glibc is forward- but
+  not backward-compatible, so a binary built on 22.04 runs on 22.04+, while one
+  built on 24.04 may not run on 22.04.
+- The build is driven by `data_forwarder_host.spec`; `scripts/build_linux_binary.sh`
+  just provisions an isolated build virtualenv (kept out of git) and invokes
+  PyInstaller against that spec.
+- The binary is large (~100 MB) because the whole Qt6 stack is embedded.
+
 ## First-time use
 
 1. Launch the application — it starts blank, with no session open and no tabs.
--- a/tools/data_forwarder_host/README.md
+++ b/tools/data_forwarder_host/README.md
@@ -1,0 +1,135 @@
+<!--
+Copyright (c) 2026 Nordic Semiconductor ASA
+SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+-->
+
+# Data Forwarder Host
+
+`data_forwarder_host` is a cross-platform desktop GUI that receives, visualises,
+records and exports sensor data forwarded from an nRF device over **UART** or
+**BLE NUS** (Nordic UART Service). Frames use the device's COBS + CBOR v1
+framing; multiple capture sessions can run side by side in their own tabs.
+
+The matching device-side firmware is the
+[`data_forwarder` sample](../../samples/data_forwarder) (`edge-ai/samples/data_forwarder`),
+which produces the COBS/CBOR `sensor-data` stream this host decodes. For BLE it
+advertises the Nordic UART Service as `nRF DataFwd`. Without that firmware
+running on a connected device there is no data to receive.
+
+## Requirements
+
+- Python **3.12** or newer.
+- Linux, Windows or macOS.
+- An nRF device running the [`data_forwarder` firmware
+  sample](../../samples/data_forwarder) (`edge-ai/samples/data_forwarder`).
+- For UART sources: permission to access the serial port (on Linux this usually
+  means membership of the `dialout` group).
+
+## Install
+
+> **Ubuntu / Debian:** Qt 6 (PySide6) needs the XCB cursor library at runtime.
+> If the app fails to start with an `xcb` platform-plugin error
+> (`Could not load the Qt platform plugin "xcb"`), install it first:
+>
+> ```bash
+> sudo apt install libxcb-cursor0
+> ```
+>
+> On a minimal install you may also need the other XCB runtime libraries
+> (`libxcb-xinerama0`, `libxkbcommon-x11-0`); these are present on a standard
+> Ubuntu desktop.
+
+From this directory (`data_forwarder_host/`), install the requirements:
+
+```bash
+pip install -r requirements.txt
+```
+
+`requirements.txt` uses `pyproject.toml` as the single source of truth, so the
+dependency list lives in exactly one place. Alternatively, install the package
+directly with pip's editable mode:
+
+```bash
+pip install -e .
+```
+
+PySide6 (Qt 6, with QtCharts) is a base dependency and is always installed — the
+application is GUI-only.
+
+## Run
+
+The application is a pure GUI — there is no command-line interface. Launch it in
+any of these equivalent ways:
+
+```bash
+data-forwarder-host          # console script (installed entry point)
+python -m data_forwarder_host
+python main.py
+```
+
+By default, source acquisition (byte reading + COBS/CBOR decode) runs in a
+separate child process so the GUI stays responsive regardless of backend load.
+Set `DFH_SINGLE_PROCESS=1` to fall back to the legacy in-GUI acquisition path.
+
+## First-time use
+
+1. Launch the application — it starts blank, with no session open and no tabs.
+2. Open `Session → New session…`.
+3. Pick a **data source**: **UART** (live serial device) or **BLE NUS** (Nordic
+   UART Service). For BLE NUS, the dialog scans for advertising devices and
+   lets you pick one to connect to (requires a working Bluetooth adapter). If
+   the BLE stream drops frames or runs slower than expected, see
+   [docs/ble_throughput_tuning.md](docs/ble_throughput_tuning.md) for an
+   ordered, OS-aware speed-up procedure.
+4. Set any other session details, then click **Create session** to open a
+   per-session tab; streaming starts automatically.
+5. Enter a recording **label**, choose an output directory, then press
+   **Record** to capture data. Recordings are buffered in RAM and dumped to
+   a CSV on **Stop**; when the output directory is left blank they are written
+   to `recordings/` next to the package. Each recording writes a `{stem}.csv`
+   data file plus a paired `{stem}.txt` metadata sidecar (host, transport,
+   timing, device session info, channels and errors). A banner shows the last
+   saved CSV with shortcuts to open the file or its containing folder.
+
+The **View** menu toggles the error and log panels, the per-tab visualisation
+panels (Combined View, Individual Channels, Channel ASCII, Decoded Frames),
+the theme (System / Light / Dark) and the log level. Session configuration can
+be saved and reopened via the **Session** menu. Live bandwidth, host metrics
+and First Failure Data Capture (FFDC) details are surfaced on each tab.
+
+## Architecture (short)
+
+The application is strictly layered. The acquisition/transport layers are fully
+**Qt-free** (no PySide6 imports), so they can run inside the headless child
+acquisition process; the data-model, session and GUI layers run in the GUI
+process and use Qt (PySide6 `QObject`/`Signal`):
+
+- `platform/` — *(Qt-free)* OS-specific adapters (serial-port enumeration,
+  access diagnostics).
+- `source/` — *(Qt-free)* byte sources: `uart` (pyserial) and `ble_nus`
+  (BLE NUS, via bleak).
+- `protocol/` — *(Qt-free)* frame decoders (COBS + CBOR v1) and the framing
+  primitives.
+- `pipeline/` — *(Qt-free)* out-of-process acquisition: a `spawn` child process
+  reads the source and decodes frames, while the GUI process drains decoded
+  envelopes over an IPC queue (toggle with `DFH_SINGLE_PROCESS`).
+- `core/` — per-session data model, rolling channel buffers, decimation, the
+  recorder, the CSV writer, the metadata sidecar and the metrics (bandwidth,
+  host, transfer, pipeline). *Uses Qt:* `data_model`, `recorder` and
+  `error_log` are `QObject`s that emit signals to the GUI; the CSV writer,
+  sidecar and the metrics dataclasses are plain Python.
+- `session/` — *(uses Qt)* `ForwardingSession` (source + decoder + I/O),
+  `SessionController` (per-tab façade) and `SessionManager` (app-wide registry).
+- `utils/` — standard paths, logging, version and small helpers (plus one
+  optional Qt-based debug-stream helper).
+- `gui/` — *(uses Qt)* PySide6 main window, menus, dialogs (`gui/dialogs/`) and
+  per-session tab widgets (`gui/widgets/`, including the QtCharts plots).
+
+## License
+
+All files carry the SPDX header:
+
+```
+Copyright (c) 2026 Nordic Semiconductor ASA
+SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+```
--- a/tools/data_forwarder_host/core/csv_writer.py
+++ b/tools/data_forwarder_host/core/csv_writer.py
@@ -1,0 +1,219 @@
+# Copyright (c) 2026 Nordic Semiconductor ASA
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+"""Fixed CSV output for a finished recording.
+
+A single CSV file in one predefined format:
+
+* Filename: ``{label}_{session}_{utc}.csv`` where ``{label}`` is the
+  user-defined recording label, ``{session}`` is the session tag, and
+  ``{utc}`` is the host UTC timestamp of the recording-start moment with
+  millisecond resolution.
+* Row 1 (header): ``device_time_ms`` followed by the N channel names and a
+  trailing ``label`` column.
+* Rows 2..M (data): the device timestamp in milliseconds followed by the N
+  channel values, in the same column order as the header, and finally the
+  user-defined recording label (constant for every row).
+
+A companion ``{stem}.txt`` sidecar holding human-readable session metadata
+(transport, device, host OS/timestamps, ``session_info``, channel layout and an
+error summary) is written next to the CSV when *metadata_text* is supplied.
+
+There is no append mode and no configurable filename pattern.
+"""
+
+from __future__ import annotations
+
+import csv
+from datetime import datetime, timezone
+from pathlib import Path
+
+from data_forwarder_host.core.recorder import Recording
+
+DEVICE_TIME_COLUMN = "device_time_ms"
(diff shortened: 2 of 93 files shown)


tools/data_forwarder_host/docs/ble_throughput_tuning.md
```diff
@@ -0,0 +1,381 @@
...
+>
+> The concepts (§1–§3, §5) are **OS-independent**. The hands-on host commands in
+> §4 are **not**, because each operating system exposes the Bluetooth stack
+> differently. Every host step is tagged with the OS it applies to:
+>
+> | Tag | Platform | Status in this tool |
+> |-----|----------|---------------------|
+> | **[Linux]**   | Linux / BlueZ        | ✅ Supported and validated. Most tunable from the host. |
+> | **[macOS]**   | macOS / CoreBluetooth | ✅ Supported. The OS auto-negotiates aggressively; almost nothing is user-tunable. |
+> | **[Windows]** | Windows / WinRT       | ⚠️ **Not yet supported / not validated.** Notes are included so you know what *would* apply, but treat them as untested. |
```
Szynkaa: > Not yet supported

is this true? or just the tweaks are not supported

tools/data_forwarder_host/docs/ble_throughput_tuning.md
```diff
@@ -0,0 +1,381 @@
...
+A connection-oriented channel has lower per-packet overhead and built-in flow
+control versus GATT notifications. Requires a new device-side transport and a
+host-side L2CAP client; note `bleak` does not currently expose L2CAP CoC, so the
+host would need a BlueZ-specific socket path.
+
+### 5.4 Explicitly drive 2M PHY + DLE from the device
+
+Call `bt_conn_le_phy_update()` and `bt_conn_le_data_len_update()` right after
+connect so the link does not rely on automatic negotiation timing. (The device
+sample already enables the buffers — `CONFIG_BT_L2CAP_TX_MTU=247`,
```
Szynkaa: Are you referring to new data-forwarder sample here? Then name it

tools/data_forwarder_host/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+The **View** menu toggles the error and log panels, the per-tab visualisation
+panels (Combined View, Individual Channels, Channel ASCII, Decoded Frames),
+the theme (System / Light / Dark) and the log level. Session configuration can
+be saved and reopened via the **Session** menu. Live bandwidth, host metrics
+and First Failure Data Capture (FFDC) details are surfaced on each tab.
+
+## Architecture (short)
+
+The application is strictly layered; the GUI-free layers carry no Qt imports:
```
mbarchie: Not exactly clear which layers are GUI-free (ofc besides `gui`). E.g. some classes in `core/` import Qt.

tools/data_forwarder_host/README.md
```diff
@@ -0,0 +1,135 @@
...
+framing; multiple capture sessions can run side by side in their own tabs.
+
+## Requirements
+
+- Python **3.12** or newer.
+- Linux, Windows or macOS.
+- For UART sources: permission to access the serial port (on Linux this usually
+  means membership of the `dialout` group).
+
+## Install
```
mbarchie, line 28: I'd mention that `libxcb-cursor0` has to be installed on Ubuntu (not sure about other deps)

tools/data_forwarder_host/pyproject.toml
```diff
@@ -0,0 +1,96 @@
...
+    "data_forwarder_host.core",
+    "data_forwarder_host.gui",
+    "data_forwarder_host.gui.dialogs",
+    "data_forwarder_host.gui.widgets",
+    "data_forwarder_host.platform",
+    "data_forwarder_host.protocol",
+    "data_forwarder_host.session",
+    "data_forwarder_host.source",
+    "data_forwarder_host.utils",
+]
```
mbarchie, line 66: `data_forwarder_host.pipeline` missing?

tools/data_forwarder_host/docs/ble_throughput_tuning.md
pioso-dev: Good catch — reworded. The host app itself is cross-platform (it uses bleak, which has a WinRT backend, and there's a wired-up WindowsPlatform adapter), so it should run on Windows; the accurate statement is that it's not validated and exposes no host-side tuning knobs — not that it's unsupported. Updated the table so only Linux/Ubuntu is marked validated, while macOS and Windows are now "Should run but not validated (untested)." Also changed the §4.2 "(not yet supported)" tag to "(not validated)." Note I downgraded macOS too, since only Ubuntu has actually been tested.

tools/data_forwarder_host/docs/ble_throughput_tuning.md
pioso-dev: Yes - it's samples/data_forwarder; I've named it explicitly in §5.4. One caveat worth flagging: that sample is currently still a stub and the cited CONFIG_BT_L2CAP_TX_MTU=247 / CONFIG_BT_BUF_ACL_RX_SIZE=251 aren't in its prj.conf yet, so I changed "already enables" -> "should enable." Once the sample ships with those configs I'll switch the wording back to "already enables."

pioso-dev (commented):
Comments resolved

tools/data_forwarder_host/core/csv_writer.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    return RecordingCsvDump(
+        recording, label=label, output_dir=output_dir, metadata_text=metadata_text
+    ).run_to_completion()
+
+
+class RecordingCsvDump:
+    """Incremental writer for a finished recording's CSV.
+
+    The whole capture is RAM/​spill-buffered while recording; at *stop* it must
```
mbarchie: There's a zero-width space in 'RAM/​spill-buffered' part

tools/data_forwarder_host/gui/dialogs/new_session_dialog.py
```diff
@@ -0,0 +1,1514 @@
...
+        )
+
+    # ------------------------------------------------------------------
+    # BLE: Bluetooth toggle + live list + connect-in-dialog
+    # ------------------------------------------------------------------
+
+    def _run_off_thread(self, fn, on_done, on_failed) -> None:
+        """Run a blocking *fn* on a worker thread; deliver result on GUI thread.
+
+        The result/​failure handlers must run on the **GUI thread** (they touch
```
mbarchie: There's a zero-width space in "reset/​failure" part

tools/data_forwarder_host/protocol/framing.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2026 Nordic Semiconductor ASA
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+"""Framing primitives.
+
+Ported verbatim from edge-ai/tools/data_forwarder_host/python_prototype/
```
mbarchie: stale reference to `edge-ai/tools/data_forwarder_host/python_prototype/`

tools/data_forwarder_host/platform/base.py
```diff
@@ -0,0 +1,141 @@
+# Copyright (c) 2026 Nordic Semiconductor ASA
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+"""Platform adapter base class and shared helpers.
+
+The Nordic / SEGGER detection heuristics (``NRF_HINTS``, ``_looks_like_nrf``,
+``_vid_pid``, ``_describe_port``, ``_short_port``) are ported verbatim from
+``edge-ai/tools/data_forwarder_host/python_prototype/simple_receiver/
```
mbarchie: stale reference to `edge-ai/tools/data_forwarder_host/python_prototype/simple_receiver`

tools/data_forwarder_host/README.md
mbarchie, line 1: Missing reference to the `data_forwarder` sample (firmware)

mbarchie (commented):
First round of review - high level code inspection + running without `data_forwarder` firmware (just starting the host app). I'll continue checking the app when I get a sample of nrf54l15tag.

PR description mentions tests, but I can't find them anywhere in the tree.

One comment about GUI - on dark theme Ubuntu GUI is poorly readable when using dark mode in Ubuntu (dark grey background + mostly black fonts). I guess "automatic color" could fix it if there is such a setting.

tools/data_forwarder_host/pyproject.toml
pioso-dev: true - I forgot to add this one

tools/data_forwarder_host/core/csv_writer.py
pioso-dev: Nice catch. There was a zero-width space (U+200B) tucked into RAM/spill-buffered — I've removed it

tools/data_forwarder_host/gui/dialogs/new_session_dialog.py
pioso-dev: Same hidden U+200B character here in result/failure — removed. I also did a quick sweep of the tree, and these two were the only spots, so we should be clear now

tools/data_forwarder_host/platform/base.py
pioso-dev: Good point, thanks. That python_prototype/simple_receiver path no longer exists, so I've removed the stale reference from both the docstring and the inline comment

tools/data_forwarder_host/protocol/framing.py
pioso-dev: The python_prototype/ tree is gone now, so I dropped the outdated "ported from …/simple_receiver.py" note from the docstring

tools/data_forwarder_host/README.md
pioso-dev: Thanks, agreed that was missing. I've added a link to the device-side sample: samples/data_forwarder in both the intro and Requirements, and noted that it advertises over BLE as nRF DataFwd and that there's no data to receive without it running

tools/data_forwarder_host/README.md
Szynkaa: I didn't have to install it, but maybe is a dependency of some other software I use

pioso-dev: @mbarchie 
Good point, thanks. I've added a note in the Install section to:
sudo apt install libxcb-cursor0
on Ubuntu/Debian (with a mention of the other XCB runtime libs for minimal installs), since that's what the xcb platform-plugin error usually needs

tools/data_forwarder_host/README.md
pioso-dev: You're absolutely right, thanks — the wording was misleading. core/ does import Qt (data_model, recorder, error_log are QObjects with signals), and so do session/ and one utils/ debug helper. I've reworded the architecture section to make the genuinely Qt-free layers explicit (platform, source, protocol, pipeline — the ones running in the headless child process) and clearly mark the layers that do use Qt

pioso-dev (commented):
@mbarchie

Thanks for the thorough first round, and for taking the time to run the host app standalone. Sounds good on picking it back up once you've got an nRF54L15 tag — the device side is the data_forwarder sample (edge-ai/samples/data_forwarder). I put the nRF54L15 tag with a battery on your desk

Tests: good catch. That line in the description was misleading for this PR, so I've fixed it. It no longer claims an in-tree pytest suite, and I also corrected the related "unit-testable headless" wording in the architecture section. Requirements and testing for the app are handled outside this PR; what's shipped here was verified through the manual end-to-end checks now listed under Verification (UART + BLE capture, multi-session tabs, recording → CSV + metadata sidecar, and GUI responsiveness under flood). If we want a dedicated automated test suite living next to the tool, I'm happy to add that as a follow-up.

Dark theme readability: you're right, and "automatic color" was a good point. Turned out there were three causes:

1) the "System" theme applied Qt's fixed light palette regardless of the OS color scheme;
2) the palette only defined a few roles, so a lot of widgets fell back to light shades in dark mode;
3) a handful of widgets hardcoded light-only colors.

Fixes: the System theme now follows the OS scheme via QStyleHints.colorScheme(), the light/dark palettes are complete (including disabled text), the app forces the Fusion style so the palette actually gets honored, and the hardcoded banner/label/status colors are now theme-aware (or picked to read fine on both). I also reworked the welcome screen's "New Session" button so it's a clear call-to-action in every mode. Next time you have it open, could you re-test on dark Ubuntu? It should be readable now in System/Light/Dark.

tools/data_forwarder_host/gui/session_tab.py
```diff
@@ -0,0 +1,200 @@
...
+        outer.addWidget(self._no_data_banner)
+
+        self._sensor_data_seen = False
+
+        # ── Main horizontal split: narrow controls | scrollable content ─
+        self._body = QSplitter(Qt.Orientation.Horizontal)
+        outer.addWidget(self._body, 1)
+
+        ctrl_panel = ControlPanel(controller)
+        ctrl_panel.setMaximumWidth(240)
```
mbarchie: IMO this could be removed or changed to a higher value. It prevents from resizing control panel, so you cannot see full values of some of the fields in "Session Configuration" and "Session Info". When I try to resize it, it shrinks (probably to the upper limit) and cannot be enlarged.

tools/data_forwarder_host/gui/widgets/control_panel.py
```diff
@@ -0,0 +1,372 @@
...
+            ("dr", "Drop Count", str(d.get("dr", "—"))),
+            ("name", "Device Name", str(d.get("name", "—"))),
+        ]
+
+        for key, label, value in field_defs:
+            field = QLineEdit()
+            field.setText(value)
+            field.setReadOnly(True)
+            self._metadata_fields[key] = field
+            self._metadata_form.addRow(f"{label}:", field)
```
mbarchie: Minor UX-related issue - these fields look like editable (like the ones in "Session Configuration"), but they're not (and ofc they shouldn't be) which may be misleading for users. It would be nice to have them looking different than the editable ones.

Another "nice-to-have" - tips when hovering over the parameters, like for "Session Configuration" items.

tools/data_forwarder_host/gui/widgets/bandwidth_details.py
```diff
@@ -0,0 +1,641 @@
...
+    "Total count",
+    "Since reset",
+    "Total bytes",
+    "Bytes since reset",
+)
+
+# Label for the window-scoped reset button. It now clears more than the
+# since-reset counters — it also zeroes each pipeline stage's drop/issue counter
+# and removes the red node-warning borders.
+CLEAR_BUTTON_LABEL = "Clear counters & node warnings"
```
mbarchie: Single ampersand is used by Qt to make "keyboard shortcuts", so it displays now as a little underscore.

```suggestion
CLEAR_BUTTON_LABEL = "Clear counters && node warnings"
```

mbarchie:
> Dark theme readability: you're right, and "automatic color" was a good point. Turned out there were three causes:
> 
> 1. the "System" theme applied Qt's fixed light palette regardless of the OS color scheme;
> 2. the palette only defined a few roles, so a lot of widgets fell back to light shades in dark mode;
> 3. a handful of widgets hardcoded light-only colors.
> 
> Fixes: the System theme now follows the OS scheme via QStyleHints.colorScheme(), the light/dark palettes are complete (including disabled text), the app forces the Fusion style so the palette actually gets honored, and the hardcoded banner/label/status colors are now theme-aware (or picked to read fine on both). I also reworked the welcome screen's "New Session" button so it's a clear call-to-action in every mode. Next time you have it open, could you re-test on dark Ubuntu? It should be readable now in System/Light/Dark.

GUI looks is now readable also on dark theme. The app looks nice!

mbarchie (commented):
Minor comments that can be handled in future PRs

mbarchie (approved):

tools/data_forwarder_host/gui/widgets/control_panel.py
pioso-dev: The read-only metadata QLineEdits are now styled to be visually distinct from the editable Session Configuration inputs: frameless (setFrame(False)), transparent/borderless background via stylesheet, and non-focusable (setFocusPolicy(NoFocus)), so they read as static labels rather than text inputs

Added a tooltip to each metadata entry in field_defs and applied it to both the value field and its row label (mirroring the setToolTip pattern used by the Session Configuration spin boxes).

tools/data_forwarder_host/gui/session_tab.py
pioso-dev: Removed the ctrl_panel.setMaximumWidth(240) constraint. The panel is now freely resizable via the splitter. The existing setStretchFactor(0, 0) still keeps it from greedily eating extra horizontal space on window resize, but the user can now drag the splitter to enlarge it and see full field values.

tools/data_forwarder_host/gui/widgets/bandwidth_details.py
pioso-dev: Applied the reviewer's suggested change — escaped the ampersand as && so Qt renders a literal & instead of a mnemonic underline.

pioso-dev (commented):
@mbarchie 
Thanks for the follow-up review and the approval! I've addressed all four minor comments in the latest push

Szynkaa (approved):

PR #113, opened 2026-06-19
nrf_edgeai_obsv — input-feature stream + mel descriptor metrics

rrusak:
This PR extends the observability library from a single output (class-probability) stream to a two-stream design, so metrics can also be computed over the extracted model input features. Each metric declares a source (PROBS or FEATURES) in its descriptor, and the core routes an update only to metrics matching the fed stream (nrf_edgeai_obsv_update vs nrf_edgeai_obsv_update_features), keeping the two cadences independent. On top of this it adds two feature metrics for ww/kws mel spectrograms: mel_energy_desc (mean/max energy, dynamic range, floor ratio, normalized against configured [p01, p99] percentiles) and mel_spectral_desc (eight scale-invariant spectral-shape rows: band ratios, centroid, spread, entropy, flatness, contrast). All metrics remain self-contained descriptors with caller-provided storage and emit a row-major uint32 histogram snapshot for the existing CDR transport. Distribution-style metrics that live in a uniform [0, 1] range no longer store bin edges and bin values in O(1) via a shared helper, while the probability-distribution metric keeps custom edges. The change also hardens the hot path: a runtime cap prevents scratch[] overflow in release builds where assert() is compiled out, and the spectral metric clamps negative inputs to keep logf/sqrtf well-defined.


include/nrf_edgeai_obsv/nrf_edgeai_obsv.h
```diff
@@ -100,6 +100,20 @@ int nrf_edgeai_obsv_reset(nrf_edgeai_obsv_ctx_t *ctx);
  */
 int nrf_edgeai_obsv_update(nrf_edgeai_obsv_ctx_t *ctx, const float *probs);
```
mbarchie: I'd rename this function - e.g. `nrf_edgeai_obsv_update_probs` - while it's still not "production ready" and API stability is not crucial at this point. IMO it's not intuitive why regular "update" function updates probabilities, not features.

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
```diff
@@ -131,6 +169,11 @@ typedef struct nrf_edgeai_obsv_metric_s {
 enum nrf_edgeai_obsv_metric_id {
 	NRF_EDGEAI_OBSV_METRIC_ID_TRANSITION_MATRIX = 2,
 	NRF_EDGEAI_OBSV_METRIC_ID_PROBS_DISTRIBUTION = 3,
+	NRF_EDGEAI_OBSV_METRIC_ID_PREDICTION_SWITCHING_RATE = 4,
+	NRF_EDGEAI_OBSV_METRIC_ID_PROBS_ENTROPY_DIST = 5,
+	NRF_EDGEAI_OBSV_METRIC_ID_PROBS_TOP2_MARGIN_DIST = 6,
+	NRF_EDGEAI_OBSV_METRIC_ID_MEL_ENERGY_DESC = 7,
+	NRF_EDGEAI_OBSV_METRIC_ID_MEL_SPECTRAL_DESC = 8,
 };
```
mbarchie, line 177: Did you think about making some separation between probs-based and feature-based metrics? Like feature-based starting from e.g. 1000? Just a thought for your consideration.

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_mel_spectral_desc.c
```diff
@@ -0,0 +1,219 @@
...
+	memset(counts, 0, sizeof(uint32_t) * NRF_EDGEAI_OBSV_MSD_NUM_ROWS * hdr->bin_num);
+}
+
+static void msd_update(const float *p_feats, uint16_t n, void *priv)
+{
+	const _nrf_obsv_msd_hdr_t *hdr = priv;
+	uint32_t *counts = msd_counts(hdr);
+	const uint8_t bins = hdr->bin_num;
+
+	assert(n <= hdr->num_features);
```
mbarchie, line 87: MSD has only an assert (which will be compiled out in release builds) while MED limits `n` to `num_features`. I think behavior could be aligned across all the metrics if no clear reason for that.

lib/nrf_edgeai_obsv/metrics/nrf_obsv_dist_binning.c
mbarchie, line 1: Every other `nrf_obsv_metric_*` file in this folder is an actual metric (has a `*_create()`, ops table, and `metric_id`), but this one is just shared binning helpers. The metric_ prefix reads like "this is a metric," and dist is easy to confuse with the real `nrf_obsv_metric_probs_distribution.c`.

Could we rename it to something like `nrf_obsv_dist_binning` (keeps the tie to the DIST_BINNING Kconfig) or `nrf_obsv_binning`? Location in metrics/ is fine since it's a private helper for the metrics.

lib/nrf_edgeai_obsv/nrf_edgeai_obsv_core.c
```diff
@@ -123,27 +123,44 @@ int nrf_edgeai_obsv_core_deregister(nrf_edgeai_obsv_core_t *p_ctx,
...
+int nrf_edgeai_obsv_core_update_features(nrf_edgeai_obsv_core_t *p_ctx, const float *p_feats,
+					 uint16_t n)
+{
+	if ((p_ctx == NULL) || (p_feats == NULL)) {
+		return -EINVAL;
+	}
+
+	dispatch(p_ctx, NRF_EDGEAI_OBSV_SOURCE_FEATURES, p_feats, n);
+
+	return 0;
```
mbarchie, line 164: Will we need to keep number of feature extractions too? (like `p_ctx->num_inferences`)

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_mel_spectral_desc.c
```diff
@@ -0,0 +1,219 @@
...
+	const float span = (float)(n - 1);
+
+	float low_ratio = clip01(low_s * inv_s);
+	float mid_ratio = clip01(mid_s * inv_s);
+	float high_ratio = clip01(high_s * inv_s);
+	float centroid_n = (n >= 2) ? clip01(centroid / span) : 0.0f;
+	float spread_n = (n >= 2) ? clip01(sqrtf(var) / (span * 0.5f)) : 0.0f;
+	float entropy_n = (n >= 2) ? clip01(ent / logf((float)n)) : 0.0f;
+	float flatness = clip01(geomean / (mean + MSD_EPS));
+	float contrast_n = clip01(((fmax - fmin) / (mean + MSD_EPS)) / (float)n);
```
mbarchie: These ones can be const too, like the other variables in the function

lib/nrf_edgeai_obsv/Kconfig
```diff
@@ -62,4 +69,124 @@ config NRF_EDGEAI_OBSV_METRIC_TRANSITION_MATRIX
...
+
+config NRF_EDGEAI_OBSV_METRIC_MEL_SPECTRAL_DESC
+	bool "Mel spectral descriptor metric"
+	select NRF_EDGEAI_OBSV_DIST_BINNING
+	help
+	  Per-frame spectral-shape statistics of the mel feature vector
+	  (low/mid/high band energy ratios, spectral centroid, spread, entropy,
+	  flatness, contrast), each accumulated into a [0, 1] histogram. All rows
+	  are scale-invariant (they divide by total energy or mean), so no
+	  amplitude calibration is needed. Uses logf/expf/sqrtf (libm). Consumes
```
mbarchie, line 178: It does use logf/expf/sqrtf quite a lot, making this metric more demanding than the other ones. I'd state it explicitly somewhere, could be also here.

include/nrf_edgeai_obsv/nrf_edgeai_obsv_core.h
```diff
@@ -128,6 +128,22 @@ int nrf_edgeai_obsv_core_deregister(nrf_edgeai_obsv_core_t *p_ctx,
  */
 int nrf_edgeai_obsv_core_update(nrf_edgeai_obsv_core_t *p_ctx, const float *p_probs);
```
mbarchie: Same comment as for `nrf_edgeai_obsv_update` - maybe worth adding `_probs` as postfix?

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_mel_energy_desc.c
```diff
@@ -0,0 +1,191 @@
...
+			floor_cnt++;
+		}
+
+		float vn = (v - hdr->scale_min) / range;
+
+		if (vn < 0.0f) {
+			vn = 0.0f;
+		} else if (vn > 1.0f) {
+			vn = 1.0f;
+		}
```
mbarchie: Is there a reason for not using `clip01()` like in MSD?

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_probs_entropy_dist.c
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+
+	float h_norm = h / logf((float)n);
+
+	if (h_norm < 0.0f) {
+		h_norm = 0.0f;
+	} else if (h_norm > 1.0f) {
+		h_norm = 1.0f;
+	}
```
mbarchie: Is there a reason for not using `clip01()` like in MSD?

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
```diff
@@ -118,6 +146,16 @@ typedef struct nrf_edgeai_obsv_metric_s {
...
 
+	/**
+	 * @brief Data stream this metric consumes (@ref nrf_edgeai_obsv_source).
+	 *
+	 * Set by the metric's @c *_create() helper: @c NRF_EDGEAI_OBSV_SOURCE_PROBS
+	 * for the probability metrics, @c NRF_EDGEAI_OBSV_SOURCE_FEATURES for the
+	 * mel descriptor metrics. The core dispatches an update to this metric only
+	 * when the fed stream matches this value.
+	 */
+	uint8_t source;
```
mbarchie: The field holds an enum value but is stored as a raw `uint8_t`, and `dispatch()` compares it as a plain `uint8_t`. Not a big deal, but so we lose the enum's "type safety", and there's no comment explaining the narrowing. Could we either declare it `enum nrf_edgeai_obsv_source source` or state that narrowing to `uint8_t` is intentional? (e.g. for struct packing - but here I think it does not help anyway)

lib/nrf_edgeai_obsv/metrics/nrf_obsv_dist_binning.c
```diff
@@ -0,0 +1,40 @@
...
+	}
+}
+
+uint8_t _dist_find_bin(const float *edges, uint8_t bin_num, float val)
+{
+	for (uint8_t b = 0; b < bin_num - 1; b++) {
+		if (val < edges[b]) {
+			return b;
+		}
+	}
```
mbarchie, line 24: Just a nit - `b` is used for indexing, while `_dist_uniform_edges` uses `i` for the same thing. I know it was already like this, it just caught my eye now when moved to new file :)

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_mel_energy_desc.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+static void med_init(const void *p_cfg, void *priv)
+{
+	const _nrf_obsv_med_hdr_t *hdr = priv;
+	uint32_t *counts = med_counts(hdr);
+
+	(void)p_cfg;
+
+	memset(counts, 0, sizeof(uint32_t) * NRF_EDGEAI_OBSV_MED_NUM_ROWS * hdr->bin_num);
```
mbarchie: Metrics in their init() functions could call clear() functions to de-duplicate "clearing" code. I think it applies to all the metrics currently implemented.

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
rrusak: I don't have a strong opinion on this matter, but since you will be the main maintainer of this library, I would rely more on your opinion :) .

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_mel_spectral_desc.c
rrusak: MED limit features to n = CONFIG_NRF_EDGEAI_OBSV_MEL_ENERGY_DESC_MAX_FEATURES because it's create buffer on stack  "float scratch[CONFIG_NRF_EDGEAI_OBSV_MEL_ENERGY_DESC_MAX_FEATURES];" so that it doesn't happen that we go beyond the limits of this buffer, and for MSD there is no such issue. In Kconfig there is CONFIG_NRF_EDGEAI_OBSV_MEL_ENERGY_DESC_MAX_FEATURES but no similar thing for MSD, for MSD only number of bins in distribution.

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
```diff
@@ -76,18 +93,29 @@ typedef struct nrf_edgeai_obsv_metric_s {
...
+	 * @param p_data Pointer to the input vector for one update: class
+	 *               probabilities (PROBS) or extracted features (FEATURES),
+	 *               per the metric's source.
+	 * @param n      Number of entries in @p p_data: @c num_classes for PROBS,
+	 *               or the length passed to @ref nrf_edgeai_obsv_update_features
+	 *               for FEATURES.
+	 * @param priv   Opaque per-instance storage pointer.
 	 */
-	void (*update)(const float *p_probs, uint16_t n, void *priv);
+	void (*update)(const float *p_data, uint16_t n, void *priv);
```
Szynkaa, line 111: I think we should replace single `update` interface with `update_probs` and `update_features`, then remove the `source` field. This lib was not released yet and it's experimental so no restrictions on breaking API

lib/nrf_edgeai_obsv/nrf_edgeai_obsv_core.c
rrusak: I believe we need this info in nrf_edgeai_obsv_model_info_t , for consistency and similarity with num_classes, I will change CDDL schema, what do you think?

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
mbarchie: On second thoughts - I think this may be not the best idea ;) Starting from e.g. 1000 will consume 2 more bytes in CBOR. Also, we actually do not know how many metrics we'll end up with. And if end up with more than 2 types of streams (shot in the dark).

I think we can keep it as it is.

lib/nrf_edgeai_obsv/nrf_edgeai_obsv_core.c
mbarchie: I guess it's a good idea

lib/nrf_edgeai_obsv/metrics/nrf_obsv_metric_mel_spectral_desc.c
mbarchie: I see, then I'm fine with it

doc/libraries/nrf_edgeai_obsv.rst
```diff
@@ -447,7 +510,8 @@ To integrate the library into your application, complete the following steps:
 #. Allocate metric storage and initialize each metric descriptor using the :c:func:`nrf_edgeai_obsv_metric_tm_create` and :c:func:`nrf_edgeai_obsv_metric_pd_create` functions.
 #. Register the metrics with the context using the :c:func:`nrf_edgeai_obsv_register` function.
 #. Bind the Memfault transport once at application startup using the :c:func:`nrf_edgeai_obsv_memfault_init` function.
-#. Call the :c:func:`nrf_edgeai_obsv_update` function after every inference.
+#. Call the :c:func:`nrf_edgeai_obsv_update_probs` function with the output probability vector after every inference.
```
Szynkaa, line 513: As I understood this is mandatory event if using only feature-source metrics. If that's true emphasize it

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
rrusak: nrf_edgeai_obsv_metric_t right now can be for features or for probs but not for both of them at the same time, so having another function pointer and one of these pointers (update_probs or update_features) will be always NULL  for specific metric seems unreasonable.

doc/libraries/nrf_edgeai_obsv.rst
rrusak: No, calling nrf_edgeai_obsv_update_probs is not mandatory if in config you are only using metrics for features, you can call nrf_edgeai_obsv_update_features only, the only drawback that num_inferences in context will be not incremented if call nrf_edgeai_obsv_update_features, maybe we need to have another counter for features vectors processed

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
Szynkaa: I agree that currently none of metrics use both inputs, but I think in theory some metric can use both. Approach with separate functions will make these functions have one job - no need to have context which type of metric this is. But I don't have strong opinion here

doc/libraries/nrf_edgeai_obsv.rst
Szynkaa: I think it depends how we plan to use the feature metrics. For example for WW and KWS model feature extraction is executed 3 time (10ms blocks) per single inference (30ms concatenated block).
Does the feature metric should be called with 10ms blocks or concatenated block. For first approach additional counter would fit, for second approach I think that inference counter is enough. How the nRF Edge AI Lib is going to handle it?

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
mbarchie: The main benefit of keeping the current approach is when it comes adding substantial amount of new sources in the future besides features and probabilities. In such case, it scales better. However I wonder if you see that's the case of observability.

Otherwise, I think the idea of removing dispatch layer is worth considering. It removes additional layer of redirection and provides possibility to have combined metrics. I could imagine metrics combining both features and probabilities, conditioning the output behavior on the input condition. For this, separate callbacks could make sense.

IMO it boils down to question what is more possible in the future - combining sources in one metric (-> let's add separate callbacks) or adding many sources in the future (-> let's keep it as it is, unless combined metrics are not impossible). That's maybe question to you @rrusak

rrusak: That’s possible in principle, but I wouldn’t want to over-engineer something that isn’t needed yet—we can always add it later :) I don't think this library will be ready for production anytime soon, which would hold us back.

doc/libraries/nrf_edgeai_obsv.rst
rrusak: For most models, the standard approach is a single feature extraction step per inference. However, KWS and wake-word models require three feature processing steps per inference because models utilize a discrete 3x40 mel-bin window rather than a sliding window. I also anticipate that future models will require multiple feature processing windows per inference.

include/nrf_edgeai_obsv/nrf_edgeai_obsv_metrics.h
mbarchie: If there are no clear predictions about future extensions of observability, I don't have strong opinion here too and IMO we can keep `source`-driven updates as-is.

One idea maybe worth considering in this area (while maintaining current way of handling metrics) would be to create separate `update_probs` and `update_features` API-facing API. This allows to drop `dispatch` function and drop `source` function argument. Finally, it also allows to keep this API unchanged regardless of the approach chosen in the future (callback-based or source-based), making it more lib-internal thing.

doc/libraries/nrf_edgeai_obsv.rst
Szynkaa: With KWS as an example, will future function for getting calculated features return single block with 40 mel-bin (so needs to be called 3 times) or full 120 bins for single inference?

rrusak: It will return single block with 40 mel-bin and update_features need to be called 3 times for one inference. Since the approach applies universally to all models and not only for WW/KWS, we settled on this design; if we run into performance issues later, we’ll look into it then.

Szynkaa (approved):

mbarchie (approved):

PR #117, opened 2026-06-24
doc: use intersphinx for NCS and Zephyr cross-references

Szynkaa:
Enable sphinx.ext.intersphinx and map the nrf and zephyr domains to the hosted NCS documentation at nrfconnectdocs.nordicsemi.com. Replace hardcoded versioned docs.nordicsemi.com URLs in links.txt with :ref: and :doc: roles across the documentation. This keeps cross-doc links aligned with the NCS version and avoids maintaining duplicate link targets locally.


mbarchie (approved):

grochu (approved):

annwoj (changes requested):
need to investigate this further

doc/conf.py
```diff
@@ -194,6 +195,18 @@
 # Consumed by the edge_ai_project_info extension (source-read handler).
 edge_ai_substitutions = SUBSTITUTIONS
 
+# -- Options for intersphinx --------------------------------------------------
+#
+# Hosted NCS inventories are published at nrfconnectdocs.nordicsemi.com.
+# See https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html
+
+_NCS_DOCS_BASE = f"https://nrfconnectdocs.nordicsemi.com/ncs/{NCS_VERSION_NUMBER}"
```
b-gent, line 203: A possible solution to firewall issues is to use the original Azure URL

```suggestion
_NCS_DOCS_BASE = f"https://ncsdoc.z6.web.core.windows.net/ncs/{NCS_VERSION_NUMBER}"
```

This may help if the issues are coming from Cloudflare.

b-gent:
These links are not created correctly or am I missing something?
image

b-gent (changes requested):

doc/conf.py
Szynkaa: will check it
Links are missing because the objects.inv download failed

b-gent (approved):

annwoj (approved):

PR #130, opened 2026-07-08
ci: add doc-required label and CODEOWNERS for docs

mbarchie:
Add a doc-required label rule and labeler workflow so PRs touching doc/**/*, *.rst or *.md files are automatically flagged for tech writer visibility. Add a CODEOWNERS file assigning those same paths to @nrfconnect/ncs-sdk-edge-ai-doc so their review can be enforced once "Require review from Code Owners" is enabled on the main branch.

Jira: NCSDK-40163

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,15 @@
+# https://github.com/actions/labeler/blob/main/README.md
+#
+# When extending this, remember that in the PR, the labeler will run against
+# the labeler.yml in main, more info:
+# https://github.com/actions/labeler/issues/12
+# This means your changes won't be tested. To test your branch, make a second
+# branch with dummy changes, and open a PR on your own fork, against the
+# first branch.
+
+"doc-required":
+- changed-files:
+  - any-glob-to-any-file:
+    - "doc/**/*"
+    - "**/*.rst"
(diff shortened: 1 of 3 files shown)


Szynkaa (approved):

.github/labeler.yml
```diff
@@ -0,0 +1,14 @@
...
+# https://github.com/actions/labeler/issues/12
+# This means your changes won't be tested. To test your branch, make a second
+# branch with dummy changes, and open a PR on your own fork, against the
+# first branch.
+
+"doc-required":
+- changed-files:
+  - any-glob-to-any-file:
+    - "doc/**/*"
+    - "**/*.rst"
```
grochu, line 14: How about the `*.md` files?

mbarchie: I based this config on the one from sdk-nrf, but it might be a good idea to add .md files too. We're not supposed to use .md anyway, but:
- we already have one README.md (for dfh)
- when .md files are in the CODEOWNERS, adding a new .md file will be tracked by tech writers too (and it can be rejected)

grochu (approved):

Szynkaa (approved):

PR #139, opened 2026-07-28
Add nRF7120 support to EdgeAI applications

erikberces:
Adds initial nrf7120dk/nrf7120/cpuapp support to three Edge AI applications: Wakeword/Keyword spotting, Person detection, and Gesture recognition. Each application gets a board-specific devicetree overlay for its sensor and the new target added to sample.yaml, so the nRF7120 DK is covered by Twister and appears in the generated table-from-sample-yaml doc tables.

One commit per application, so each can be reviewed or reverted independently.

Tickets: NCSDK-40104 NCSDK-40105 NCSDK-40106


applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/prj.conf
```diff
@@ -0,0 +1,29 @@
...
+CONFIG_GPIO=y
+
+# nRF Axon
+CONFIG_NRF_AXON_INTERLAYER_BUFFER_SIZE=512
+CONFIG_NRF_EDGEAI_GESTURE_RECOGNITION_MODEL_AXON=y
+
+# Enable sensor, button and LED libraries
+CONFIG_SPI=y
+CONFIG_SENSOR=y
+
```
PavelVPV: I don't see any LED lib is enabled here. Is the comment incorrect?

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/prj.conf
```diff
@@ -0,0 +1,29 @@
...
+# No Bluetooth in this minimal bring-up configuration
+CONFIG_BLE_MODE_NONE=y
+CONFIG_BT=n
```
PavelVPV: I guess we need a follow-up task to replicate behavior for other boards like explained here:
https://github.com/nrfconnect/sdk-edge-ai/blob/fff107482289c772275e0baebce491f1a3f8fa57/applications/gesture_recognition/README.rst?plain=1#L220-L242

And/or have a discussion what we should/can do from Wi-Fi perspective.

applications/ww_kws/README.rst
PavelVPV, line 1: I guess we need to create a follow-up task to port this sample to Wi-Fi to support observability features:
https://github.com/nrfconnect/sdk-edge-ai/blob/fff107482289c772275e0baebce491f1a3f8fa57/applications/ww_kws/README.rst?plain=1#L183-L184

PavelVPV, line 1: Do you know if `UART30` is correct one: https://github.com/nrfconnect/sdk-edge-ai/blob/fff107482289c772275e0baebce491f1a3f8fa57/applications/ww_kws/README.rst?plain=1#L205?

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/app.overlay
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+&global_peripherals {
+	axon: axon@56000 {
+		compatible = "nordic,axon";
+		reg = <0x56000 0x1000>;
+		interrupts = <86 NRF_DEFAULT_IRQ_PRIORITY>;
+		status = "okay";
+	};
+};
```
PavelVPV, line 14: @rob-robinson-14 , I guess axon declaration will be defined in some `.dtsi` file at SoC-level, right?

applications/gesture_recognition/README.rst
PavelVPV, line 1: There are several sections that needs to be aligned with nRF7102 (`Development kits`, `Button and LEDs`, etc.). Create a follow-up task (unless this can be done now)?

applications/gesture_recognition/src/nrf_edgeai_generated/nrf7120dk/Axon/nrf_edgeai_user_model.c
PavelVPV, line 1: I know these files are not something new that this PR introduces in the sample, but this is a real gap in readme IMHO. We commit generated model artifacts under `src/nrf_edgeai_generated/<board>/<model>/` for every supported platform (`thingy53`, `nrf54l15tag`, `nrf54lm20dk`, and now `nrf7120dk`), but the README never explains what they are or how to reproduce them. The only trace is inside the generated files themselves (a timestamp, a solution ID, a compiler version) — not enough for someone to regenerate or update them later. Since this PR touches the nrf7120dk case, this is a good point to close that gap for all platforms at once, rather than adding a fourth undocumented copy.

I proposed addition to `applications/gesture_recognition/README.rst`, e.g. under a new `Generated model files` section:

```
Generated model files
======================

The ``src/nrf_edgeai_generated/<board>/<model>/`` directories contain files
generated by the `Nordic Edge AI Lab`_ and are committed to the
repository so the application builds without requiring access to the Lab.

Each ``<board>/<model>/`` directory corresponds to one exported model and
contains:

* ``nrf_edgeai_user_model.c`` / ``.h`` -- generated model runtime code.
* ``nrf_edgeai_user_model_<model>.h`` -- model weights/configuration.
* ``nrf_edgeai_user_types.h`` -- shared generated type definitions.
* ``prj_example.conf`` -- Kconfig fragment matching the exported model.

Regenerating these files
-------------------------

To regenerate the files for a given board/model:

1. Open the corresponding solution in the Nordic Edge AI Lab
   (solution ID: see ``EDGEAI_LAB_SOLUTION_ID_STR`` in the existing
   ``nrf_edgeai_user_model.c``).
2. Export/generate the model for the target board using Edge AI Lab
   version <X.Y.Z> / Neural Network Compiler version <X.Y.Z>.
3. Replace the contents of the corresponding
   ``src/nrf_edgeai_generated/<board>/<model>/`` directory with the
   exported output.

Do not hand-edit files under ``nrf_edgeai_generated/`` -- any manual change
will be lost the next time the model is regenerated.
```

PavelVPV (approved):
Good work, thanks @erikberces! I have few really minor comments/questions. Please address those that make sense.

I also don't know the process for release-notes update. Need some maintainer of the repo to review this.

applications/person_detection/boards/nrf7120dk_nrf7120_cpuapp.overlay
```diff
@@ -0,0 +1,56 @@
...
+		compatible = "nordic,axon";
+		reg = <0x56000 0x1000>;
+		interrupts = <86 NRF_DEFAULT_IRQ_PRIORITY>;
+		status = "okay";
+	};
+};
+
+&axon {
+	status = "okay";
+};
```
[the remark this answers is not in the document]
erikberces: Removed the duplicate &axon { status = "okay"; }; block

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/prj.conf
erikberces: You're right, no LED lib is enabled there. I fixed the description.

applications/gesture_recognition/src/nrf_edgeai_generated/nrf7120dk/Axon/README.md
```diff
@@ -0,0 +1 @@
+# This is directory for Nordic EdgeAI Lab solution generated files
```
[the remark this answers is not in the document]
erikberces: This file is generated from Edge AI Lab. Shouldn't be changed by hand.

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/prj.conf
erikberces: I created the follow-up task in JIRA as you suggested: 
https://nordicsemi.atlassian.net/browse/NCSDK-40628

applications/ww_kws/README.rst
erikberces: Yes. On both kits the console/shell is a separate instance from the control output: on the nRF7120 DK zephyr,console and zephyr,shell-uart are &uart20 (from nrf7120_cpuapp_common.dtsi), and the overlay sets ncs,control-output-uart = &uart30, same as the nRF54LM20 DK overlay.

applications/gesture_recognition/src/nrf_edgeai_generated/nrf7120dk/Axon/nrf_edgeai_user_model.c
erikberces: You are right, I added suggested section to README file to describe the generated files.

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/app.overlay
erikberces: I can update it once it is defined in .dtsi at SoC-level.

applications/ww_kws/README.rst
erikberces: I created the follow-up task in JIRA as you suggested:
https://nordicsemi.atlassian.net/browse/NCSDK-40669

applications/gesture_recognition/README.rst
erikberces: I created the follow-up task in JIRA as you suggested:
https://nordicsemi.atlassian.net/browse/NCSDK-40670

applications/ww_kws/README.rst
rob-robinson-14: I don't think we should be using uart30 here. Pins for the uarts on the DK for nrf7120 are all on port 1. uart30 cannot use pins on port 1 (since its lp domain). So with this setup you force someone to user external FTDI to uart. FLPR and TF-M make use of uart21 for this reason.

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/app.overlay
rob-robinson-14: Yes we need to add soc / board support for nRF7120e, MDK for this has just been released so work is no longer blocked.

Note axons is only available on 7120e, not 7120 so overlay file name should be updated to reflect this but this may cause CI errors since the board doesn't exist yet.

We are hoping to do this work in next few weeks, but it may take some time to get this into sdk-nrf. I would propose that this PR stays unmerged until 7120e support exists - Just have it ready to go with appropriate file name.

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/app.overlay
```diff
@@ -0,0 +1,53 @@
...
+
+&pinctrl {
```
rob-robinson-14, line 16: I would avoid spi21 since uart21 is used for flpr / ns (and as I recommended - for additional console output in this sample).

applications/gesture_recognition/sample.yaml
```diff
@@ -200,3 +200,11 @@ tests:
       - nrf54lm20dk/nrf54lm20b/cpuapp
     extra_args:
       - gesture_recognition_SHIELD=pca63568
+  # Minimal no-BLE Axon bring-up on the nRF7120 DK (single image, no MCUboot).
+  # BLE_MODE_NONE / BT=n / MCUboot=n come from configuration/nrf7120dk_nrf7120_cpuapp.
+  applications.gesture_recognition.none.nrf7120:
```
rob-robinson-14: This uses the common regex - which references bootloader and ble? Even if build_only I would still expect sample to work / pass if a user tests it.

Does 7120 need to be in title name here?

Expectation is that bootloader and bluetooth are supported - why do they need to be disabled? I guess because west point to old version of sdk-nrf in west - are we expecting this to chage?

applications/ww_kws/boards/nrf7120dk_nrf7120_cpuapp.overlay
```diff
@@ -0,0 +1,41 @@
...
+};
+
+&uart30 {
+	status = "okay";
+};
+
+&pinctrl {
+	pdm20_default_alt: pdm20_default_alt {
+		group1 {
+			psels = <NRF_PSEL(PDM_CLK, 1, 8)>,
```
rob-robinson-14: pins 1.8 and 1.9 are reserved for onboard uart. 

Avoid on P1: 0,1,2,3,8,9,10,11

rob-robinson-14 (changes requested):

rob-robinson-14:
Given this is first addition on 7120e - should also update changelog to reflect this.

applications/ww_kws/README.rst
PavelVPV: To confirm the said above:
* UART20 is for logging, therefore the app should not use it to not mix logs with model output
* UART30 lives in LP-domain, which is bound to P0, not to P1 pins wired to the DK's onboard debugger. So it can't be read out-of-the-box without an external UART adapter
* UART21 is a good candidate for now. But we need to keep in mind that if we target to /ns variant, then we need to disable TF-M logs here and release this uart for non-secure domain. And also move from SPI21 as @rob-robinson-14 suggested in the other comment.

rob-robinson-14: Agreed - note this is similar to how 54lm20 uses uart30 (i.e. for TFM / FLPR). Although there is no ns support in edge-ai samples to date for 54lm20 (or 7120)

applications/ww_kws/boards/nrf7120dk_nrf7120_cpuapp.overlay
erikberces: I used pins 1.4 and 1.5 instead.

applications/gesture_recognition/sample.yaml
erikberces: I updated the app for nRF7120 to support bootloader and bluetooth and updated the sample.yaml. The update also removed the 7120 from the title name.

applications/ww_kws/README.rst
erikberces: Unfortunately, the addon is using old version of zephyr/sdk-nrf repos and it doesn't have the updated .dts files for nRF7120 that you did in commit https://github.com/zephyrproject-rtos/zephyr/commit/93ad08b1a77f0345355b3919a547d5ffc6ee253d . I would need to do additional changes to make it buildable. I would keep using uart30 for now and I will prepare update to match current .dts files later under ticket https://nordicsemi.atlassian.net/browse/NCSDK-40756 .

erikberces:
> Given this is first addition on 7120e - should also update changelog to reflect this.

I can't find any changelog file where they would store changes for next 2.3.0 release.

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/app.overlay
erikberces: The usage of spi will be updated under https://nordicsemi.atlassian.net/browse/NCSDK-40756

rob-robinson-14: Okay have added a comment in jira - would recommend being specific about changes required, else when we come back to the work they will be forgotten about.

Not sure why this can't just be updated form 21 to 22 now? No pin conflicts on port 3.

erikberces: I updated the SPI to 22, you are right, there is no blocker.

annwoj (approved):

PavelVPV (approved):

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/app.overlay
```diff
@@ -0,0 +1,53 @@
...
+		};
+	};
+};
+
+&gpio3 {
+	status = "okay";
+};
+
+&spi22 {
+	compatible = "nordic,nrf-spim";
```
rob-robinson-14: redundant since already defined in .dtsi.

applications/gesture_recognition/README.rst
```diff
@@ -204,6 +210,40 @@ The application supports two execution backends:
...
+=====================
+
+The :file:`src/nrf_edgeai_generated/` directory contains model files exported from the `Nordic Edge AI Lab`_.
+These files are committed to the repository so that the application builds without access to the Lab.
+
+The application selects the model directory at build time based on the board name and the selected backend:
+
+* :file:`src/nrf_edgeai_generated/<board>/Axon/` - Used when the ``CONFIG_NRF_EDGEAI_GESTURE_RECOGNITION_MODEL_AXON`` Kconfig option is enabled.
+* :file:`src/nrf_edgeai_generated/<board>/Neuton/` - Used for the Neuton backend on boards that provide both backends.
+* :file:`src/nrf_edgeai_generated/<board>/` - Used for the Neuton backend on boards that provide only this backend.
```
rob-robinson-14, line 223: Is neuton support for 7120 expected to be added at a later date? No Neuton build folder?

applications/gesture_recognition/README.rst
```diff
@@ -204,6 +210,40 @@ The application supports two execution backends:
...
+To regenerate the files for a board, complete the following steps:
+
+1. Open the corresponding solution in the `Nordic Edge AI Lab`_.
+   The solution identifier is stored in the ``EDGEAI_LAB_SOLUTION_ID_STR`` macro in the existing :file:`nrf_edgeai_user_model.c` file.
+#. Export the model for the target board and backend.
+   For Axon models, the Axon Neural Network Compiler version used for the existing files is recorded in the header comment of the :file:`nrf_edgeai_user_model_axon.h` file.
+#. Replace the contents of the model directory with the exported output.
+
+.. note::
+   Do not edit the files in the :file:`src/nrf_edgeai_generated/` directory manually, because the changes are lost when the model is regenerated.
```
rob-robinson-14: Does this not contradict: `#. Replace the contents of the model directory with the exported output.`

Or is okay to replace whole files? Think would benefit from being a little more explicit IMO.

applications/person_detection/boards/nrf7120dk_nrf7120_cpuapp.overlay
```diff
@@ -0,0 +1,52 @@
...
+				<NRF_PSEL(SPIM_MISO, 1, 13)>,
+				<NRF_PSEL(SPIM_MOSI, 1, 14)>;
+			bias-pull-down;
+			low-power-enable;
+		};
+	};
+};
+
+&spi22 {
+	compatible = "nordic,nrf-spim";
```
rob-robinson-14: same as above

applications/person_detection/boards/nrf7120dk_nrf7120_cpuapp.overlay
```diff
@@ -0,0 +1,52 @@
...
+	pinctrl-1 = <&spi22_sleep>;
+	pinctrl-names = "default", "sleep";
+	overrun-character = <0x00>;
+	zephyr,pm-device-runtime-auto;
+	cs-gpios = <&gpio0 4 GPIO_ACTIVE_LOW>;
+
+	arducam_mega: arducam-mega@0 {
+		compatible = "nordic,arducam-mega";
+		reg = <0>;
+		spi-max-frequency = <8000000>;
```
rob-robinson-14: be consistent across overlays. 


```suggestion
		spi-max-frequency = <DT_FREQ_M(8)>;
```

applications/ww_kws/README.rst
rob-robinson-14: IMO easier to do the work now around uart30, add the pin overrides in the overlay - as you have been doing for SPI configurations. Means you can tests with MalagoB and then immediately test on DK without any rework. But happy for it to be a follow up PR so this one is closed off.

rob-robinson-14 (commented):
Few nits and flagged some future work that is required but perhaps out of scope for this PR.

applications/gesture_recognition/configuration/nrf7120dk_nrf7120_cpuapp/app.overlay
erikberces: Removed the redundant line.

applications/person_detection/boards/nrf7120dk_nrf7120_cpuapp.overlay
erikberces: Removed the redundant line.

applications/person_detection/boards/nrf7120dk_nrf7120_cpuapp.overlay
erikberces: Changed to match the gesture recognition overlay.

applications/gesture_recognition/README.rst
erikberces: I agree it could be more clear. I tried to rephrase so it is not confusing.

applications/gesture_recognition/README.rst
erikberces: I will add the Neuton support for nRf7120 later under ticket [NCSDK-40833](https://nordicsemi.atlassian.net/browse/NCSDK-40833). For now I focused more on Axon.

rob-robinson-14 (approved):

applications/gesture_recognition/README.rst
PavelVPV: Remember that we agreed on focusing on Axon for now: https://nordicsemi.atlassian.net/browse/WZN-9996?focusedCommentId=2316935

PR #143, opened 2026-08-17
doc: nrf_edgeai_obsv: add top-level model observability lifecycle doc

mbarchie:
Add a conceptual overview page under Integrations explaining what model observability is and how it fits into the full Edge AI model lifecycle (creation, training, integration, deployment, observing, retraining, and updating). Move the conceptual overview prose out of nrf_edgeai_obsv.rst into the new page, leaving the library page focused on technical reference and cross-linking to the new overview. Illustrate the lifecycle with a Graphviz cycle diagram and a Mermaid swimlane-beta diagram.

Jira: NCSDK-40468

### DIFF
--- a/doc/integrations.rst
+++ b/doc/integrations.rst
@@ -4,7 +4,7 @@ Integrations
 ############
 
 The |NCS| integration in |EAI| allows using |NCS| together with third-party software components.
-The following user guides describe the available integrations:
+The following user guides describe the available integrations.
 
 .. toctree::
    :maxdepth: 1
--- a/doc/libraries/nrf_edgeai_obsv.rst
+++ b/doc/libraries/nrf_edgeai_obsv.rst
@@ -14,26 +14,21 @@ It works with any inference engine that produces a probability vector, including
 Overview
 ********
 
-Deployed Edge AI models behave differently in production then in controlled environments.
-Class distributions shift, transitions between predicted labels change over time, and confidence scores drift as conditions change.
-The observability module gives you a structured way to capture these statistics on-device and send them off-device for analysis.
+For a conceptual introduction to model observability, what it enables, and how it fits into the wider Edge AI model lifecycle, see :ref:`model_observability`.
 
 Metrics are driven by two input streams.
 Output metrics consume the model's class-probability vector passed to :c:func:`nrf_edgeai_obsv_update_probs`, while input-feature metrics consume the extracted feature vector fed to the model and passed to :c:func:`nrf_edgeai_obsv_update_features`.
 Each metric declares which stream it consumes through its ``source`` field, and the library routes every update only to the metrics that match.
 
-Collected data enables the following:
-
-* Model quality monitoring - Allows tracking whether prediction confidence and class frequencies stay within expected bounds after deployment.
-* Dataset collection guidance - Helps identify which classes are under-represented or confused in the field, and target data collection efforts accordingly.
-* Retraining triggers - Allows detecting distribution shift early and decide when a model update is needed before accuracy degrades noticeably.
-* A/B testing - Allows comparing metric snapshots from devices running different model versions to evaluate improvements in production conditions.
+.. note::
+   |EAILib| does not currently expose the extracted feature vector through its API, so input-feature metrics are not yet available for models integrated through it.
+   Support for capturing this data from |EAILib| is planned for a future release.
 
 The module is organized as three cooperating layers:
 
-* Core (:file:`lib/nrf_edgeai_obsv/`) - A portable, mutex-free state machine that accumulates metric counters as inference results arrive.
+* Core (:file:`lib/nrf_edgeai_obsv/nrf_edgeai_obsv_core.c`) - A portable, mutex-free state machine that accumulates metric counters as inference results arrive.
   It has no Zephyr RTOS dependency and you can use it in bare-metal environments, other RTOSes, or host-side test builds.
-* Zephyr wrapper (:file:`lib/nrf_edgeai_obsv/`) - Wraps the core in a mutex-protected context so that multiple threads can feed inferences and trigger encoding without data races, and integrates the library into the Zephyr build system (CMake, Kconfig, logging).
+* Zephyr wrapper (:file:`lib/nrf_edgeai_obsv/nrf_edgeai_obsv.c`) - Wraps the core in a mutex-protected context so that multiple threads can feed inferences and trigger encoding without data races, and integrates the library into the Zephyr build system (CMake, Kconfig, logging).
 * Memfault CDR transport (:file:`lib/nrf_edgeai_obsv_memfault/`) - Encodes the accumulated metric snapshots as a CBOR blob and stages them as a `Memfault Custom Data Recording`_ (CDR) that the Memfault SDK packetizer uploads on the next transport drain cycle.
   For Memfault Kconfig, keys, and transports in |NCS|, see :ref:`nrf:mod_memfault`.
 
@@ -46,7 +41,6 @@ The module is organized as three cooperating layers:
    skinparam defaultTextAlignment center
    skinparam ArrowColor #0077C8
    skinparam ArrowThickness 1
-   skinparam linetype ortho
    skinparam componentStyle rectangle
 
    skinparam component {
@@ -62,23 +56,29 @@ The module is organized as three cooperating layers:
    }
 
    together {
-     component "Application" as App
+     component "Application" as App #D9E1E2;line:768692;text:333F48
      component "nrf_edgeai_obsv\n(Zephyr wrapper + core)" as Obsv
      component "Metrics\n(e.g. probability distribution)" as Metrics
    }
 
-   component "nrf_edgeai_obsv_memfault\n(Memfault CDR transport)" as MfltTransport
-   component "Memfault SDK" as MfltSDK
-   cloud "nRF Cloud\n(Memfault)" as Cloud
-   component "Monitoring tool\n(dashboard / ML pipeline)" as Dashboard
+   component "nrf_edgeai_obsv_memfault\n(Memfault CDR transport)" as MfltTransport #0033A0;line:0033A0;text:FFFFFF
+   component "Memfault SDK" as MfltSDK #0033A0;line:0033A0;text:FFFFFF
+   cloud "nRF Cloud\n(Memfault)" as Cloud #0033A0;line:0033A0;text:FFFFFF
+   component "Monitoring tool\n(dashboard / ML pipeline)" as Dashboard #0033A0;line:0033A0;text:FFFFFF
 
-   App -right-> Obsv : inference results\n(class probabilities)
-   Obsv -down-> Metrics : accumulate counters
-   App -down-> MfltTransport : trigger collect
-   MfltTransport -up-> Obsv : encode metrics as CBOR
-   MfltTransport -right-> MfltSDK : stage CDR
-   MfltSDK -right-> Cloud : upload via BLE or HTTP
-   Cloud -right-> Dashboard : fetch CDR\n(REST API)
+   App -right-> Obsv : inference results\n(class probabilities) [1]
+   Obsv -down-> Metrics : accumulate counters [2]
+   App -down-> MfltTransport : trigger collect [3]
+   MfltTransport -up-> Obsv : encode metrics as CBOR [4]
+   MfltTransport -right-> MfltSDK : stage CDR [5]
+   MfltSDK -right-> Cloud : upload via BLE or HTTP [6]
+   Cloud -right-> Dashboard : fetch CDR\n(REST API) [7]
+
+   legend right
+     Two independent flows start at "Application":
+     1-2 run on every inference.
+     3-7 run periodically, or on demand, to drain and upload the accumulated metrics.
+   endlegend
 
 The following diagram shows the detailed call sequence between the observability layers, the application, and the Memfault SDK.
 
--- a/doc/model_lifecycle.rst
+++ b/doc/model_lifecycle.rst
@@ -1,0 +1,301 @@
+.. _model_observability:
+.. _model_lifecycle:
+
+Model lifecycle and observability
+#################################
+
+.. contents::
+   :local:
+   :depth: 2
+
+Model observability closes the loop between a neural network model running on a device in the field and the tools you used to create it.
+This page provides an overview of model observability, the production issues it can help you investigate, and where it fits in the Edge AI model lifecycle — from model creation through deployment, field monitoring, analysis, and remediation.
+
+To integrate observability into your application, see :ref:`nrf_edgeai_obsv_lib`.
+
+Overview
+********
+
+Deployed Edge AI models behave differently in production than in controlled environments.
+Class distributions shift, transitions between predicted labels change over time, and confidence scores drift as conditions change.
+Model observability gives you a structured way to capture these statistics on-device and send them off-device for analysis, so you can tell whether a deployed model is still performing as expected.
+
+The |EAI| implements model observability through the :ref:`nrf_edgeai_obsv_lib`.
+The library works with any inference engine that exposes a class-probability vector, an input-feature vector, or both — depending on which built-in or custom metrics you enable — including the :ref:`nrf_edgeai_lib`, :ref:`Axon NPU <lib_axon>`, and `Edge Impulse`_ deployments.
+
+.. note::
+   Model observability is designed primarily around |EAILib|, which provides the most seamless integration.
+   To get the most out of model observability, consider using |EAILib| for your models.
+
+Use cases
+*********
+
+Collected observability data enables the following:
+
+* Model quality monitoring — Tracks if prediction confidence and class frequencies stay within expected bounds after deployment.
+* Dataset collection guidance — Identifies which classes are under-represented or confused in the field and targets data collection accordingly.
+* Retraining triggers — Detects distribution shift early and decides when a model update is needed before accuracy degrades noticeably.
+* A/B testing — Compares metric snapshots from devices running different model versions in production conditions.
+* Hardware and environment health signals — Helps investigate whether a shift is model drift, a degrading sensor, or an unexpected deployment environment.
+* Software regression detection — Correlates a sudden metric shift with an application update timeline, or rules out post-processing defects when the product misbehaves but metrics stay within bounds.
+
+.. _model_observability_cycle_diagram:
+
+The Edge AI model lifecycle
+***************************
+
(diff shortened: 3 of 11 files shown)


doc/integrations.rst
```diff
@@ -4,13 +4,15 @@ Integrations
 ############
 
 The |NCS| integration in |EAI| allows using |NCS| together with third-party software components.
+See :ref:`model_observability` for a conceptual overview of how a model moves from creation to deployment, observation, and retraining.
```
Szynkaa: referred model observability is just one part of deploy-observe-retrain flow, so maybe we should rephrase it?

doc/integrations/model_observability.rst
```diff
@@ -0,0 +1,252 @@
...
+           reviseHw(["Revise hardware, calibration, or product environment (9)"])
+       end
+
+       train --> integrateModel
+       configureObsv --> deployProduct
+       uploadMetrics --> analyzeMetrics
+
+       rootCause -- model --> retrainModel --> updateModel --> deployUpdate
+       rootCause -- software pipeline --> fixCode --> deployUpdate
+       rootCause -- hardware or environment --> reviseHw
```
Szynkaa: this path takes a strange path on diagram

doc/integrations/model_observability.rst
```diff
@@ -0,0 +1,252 @@
...
+     n3 [label="Deployment", fillcolor="#3AA9F0", fontcolor="#FFFFFF"];
+     n4 [label="Observing\nproduction behavior", fillcolor="#0077C8", fontcolor="#FFFFFF"];
+     n5 [label="Analyzing\ncollected data", fillcolor="#044570", fontcolor="#FFFFFF"];
+
+     n1 -> n2 -> n3 -> n4 -> n5 -> n1;
+   }
+
+The following diagram breaks this cycle down further, showing where each stage fits relative to the |EAI| tooling, and the branches that analysis can take.
+
+.. mermaid::
```
Szynkaa: this diagram is very big on my screen, we should set a limit on it's maximum size

doc/integrations/model_observability.rst
```diff
@@ -0,0 +1,252 @@
...
+       rootCause -- software pipeline --> fixCode --> deployUpdate
+       rootCause -- hardware or environment --> reviseHw
+       issueDetected -- no --> continueMonitoring
+
+      classDef decision fill:#8CCBFF,stroke:#0077C8,stroke-width:1.5px,color:#333F48;
+      classDef terminal fill:#044570,stroke:#333F48,stroke-width:1.5px,color:#ffffff;
+       class issueDetected,rootCause decision
+       class startNode,continueMonitoring,deployUpdate,reviseHw terminal
+
+The numbered stages map to the following parts of the model lifecycle.
```
Szynkaa: I don't understand mapping between numbers (1-9) and model lifecycle (5 stages)

doc/integrations/model_observability.rst
```diff
@@ -0,0 +1,252 @@
+.. _model_observability:
+
+Model observability
```
Szynkaa: page title is observavility but then we have lifecycle, I understand that this is modified lifecycle for observability but still looks strange to me

in general I think some part of lifecycle should be in overwiew.rst of whole solution

mbarchie: Rather than moving lifecycle content into `overview.rst` (which is scoped to comparing training toolchains/runtimes and isn't really about what happens after deployment), I'd rather rename this page to e.g. "Model lifecycle and observability" so the title matches its actual scope. It's difficult to explain what observability is for without the lifecycle context.

That would also make the `integrations.rst` cross-reference accurate without rewording it.

doc/integrations/model_observability.rst
mbarchie: In a separate [commit](https://github.com/nrfconnect/sdk-edge-ai/pull/143/changes/19a65e159eb7a29b6794a432c155830b0fce37a4) I restructured the page, mapping between both diagrams and text should be clearer in this case

doc/integrations/model_observability.rst
```diff
@@ -0,0 +1,256 @@
...
+
+The |EAI| provides model observability through the :ref:`nrf_edgeai_obsv_lib`, which works with any inference engine that exposes a class-probability vector, an input-feature vector, or both, depending on which built-in or custom metrics you enable, including the :ref:`nrf_edgeai_lib`, :ref:`Axon NPU <lib_axon>`, and `Edge Impulse`_ deployments.
+
+.. _model_observability_cycle_diagram:
+
+The Edge AI model lifecycle
+***************************
+
+A model does not stop evolving once it is deployed.
+It moves through a cycle of creation, integration, deployment, observation, and analysis, and observability is what tells you when to start the cycle again.
```
Szynkaa:
```suggestion
It moves through a cycle of creation, integration, deployment, observation, and analysis.
Observability is what tells you when to start the cycle again.
```

doc/integrations.rst
```diff
@@ -14,3 +14,4 @@ The following user guides describe the available integrations:
    integrations/axon
    integrations/nrf_edgeai
    integrations/edge_impulse
+   integrations/model_observability
```
Szynkaa: in nrf it would be in "Application Development" but we lacks such section so I think integrations is good place for this, @annwoj to confirm

doc/integrations/model_observability.rst
```diff
@@ -0,0 +1,256 @@
...
+     n1 [label="Model training/retraining (1, 6a)", fillcolor="#CFE8FF", fontcolor="#333F48"];
+     n2 [label="Application\nintegration (2)", fillcolor="#8CCBFF", fontcolor="#333F48"];
+     n3 [label="Deployment (3)", fillcolor="#3AA9F0", fontcolor="#FFFFFF"];
+     n4 [label="Observing\nproduction behavior (4)", fillcolor="#0077C8", fontcolor="#FFFFFF"];
+     n5 [label="Analyzing\ncollected data (5)", fillcolor="#044570", fontcolor="#FFFFFF"];
+
+     n1 -> n2 -> n3 -> n4 -> n5 -> n1;
+   }
+
+The following diagram breaks this model lifecycle down into the same numbered steps, plus the |EAI| tooling behind each one, and adds steps 6b and 6c for the cases where the analysis in step 5 points outside the model.
```
Szynkaa: > plus the |EAI| tooling behind each one

there is no any tool name on the diagram

mbarchie:
@annwoj - I'm adding you to clarify placement of the new page within documentation - if it should go under "Integrations".

I'll keep PR in draft until I get input also from @rrusak and @wbober about the content - then full doc review can be done.

doc/integrations/model_observability.rst
```diff
@@ -0,0 +1,256 @@
...
+
+Not every quality issue that observability data reveals traces back to the model itself.
+A sustained drop in confidence or a rise in ambiguous predictions can also come from degrading or poor-quality hardware, for example a microphone that no longer picks up enough signal, or from a deployment environment the model was not trained for, for example a location with more background noise than the training data.
+These cases call for revising the hardware or the product's environment, not retraining (see :ref:`model_observability_hardware_environment`).
+
+Observability data can also point to a defect in your own application code rather than in the model or the hardware.
+Because the library captures the probability and feature vectors right at inference time, a bug in the pre-processing stage feeding data into the model, for example incorrect input windowing, changes what the metrics see and can distort them in ways that look like model drift, typically as a sudden change that coincides with a firmware update rather than a slow trend.
```
rrusak: The library includes a number of checks to prevent an incorrect number of input data samples. 
But if the data format itself suddenly changes then yes, for example, suppose the accelerometer was set to a full-scale range of 2g, but then suddenly switched to 16g

doc/integrations/model_observability.rst
```diff
@@ -0,0 +1,256 @@
...
+A defect in post-processing logic downstream of inference, for example a decision threshold applied to the wrong class index, does not distort these metrics directly, but it can surface indirectly, when the product visibly misbehaves in the field while the metrics stay within expected bounds (see :ref:`model_observability_software_pipeline`).
+
+Observability data also reflects how people actually use the product, which does not always match the assumptions made while collecting training data.
+For example, a wearable held or worn at unexpected angles, or a keyword spotting model triggered with phrases or accents outside its training vocabulary, can show up as class confusion or a higher switching rate.
```
rrusak: **or a keyword spotting model triggered with phrases or accents outside its training vocabulary**

Unfortunately, we cannot determine this using our observabilities, to do so, we would need to know the ground truth label on the device. What we can diagnose:
- The person is saying the keyword, but it never lands as a confident, held detection.
- Whether a command ever worked and used in a session.
- How consistent the model's decisions are.
- How confident vs. hesitant the model is
- How busy vs. idle each device is.
- Which commands the product confuses with each other.

doc/libraries/nrf_edgeai_obsv.rst
```diff
@@ -14,21 +14,12 @@ It works with any inference engine that produces a probability vector, including
...
-Collected data enables the following:
-
-* Model quality monitoring - Allows tracking whether prediction confidence and class frequencies stay within expected bounds after deployment.
-* Dataset collection guidance - Helps identify which classes are under-represented or confused in the field, and target data collection efforts accordingly.
-* Retraining triggers - Allows detecting distribution shift early and decide when a model update is needed before accuracy degrades noticeably.
-* A/B testing - Allows comparing metric snapshots from devices running different model versions to evaluate improvements in production conditions.
-
 The module is organized as three cooperating layers:
 
 * Core (:file:`lib/nrf_edgeai_obsv/`) - A portable, mutex-free state machine that accumulates metric counters as inference results arrive.
```
rrusak: lib/nrf_edgeai_obsv/nrf_edgeai_obsv_core.c ?

doc/libraries/nrf_edgeai_obsv.rst
rrusak: lib/nrf_edgeai_obsv/nrf_edgeai_obsv.c ?

rrusak, line 36: Just a point I’d like to highlight: this diagram is a bit hard to read because it’s unclear where the process begins and ends :)

rrusak, line 331: I believe it worth to add math formula for each metrics we calculate. For Mel spectral descriptor the same

doc/integrations/model_observability.rst
mbarchie: I think we should assume that observability module can be used with other runtimes than Edge AI Lib, which may or may not include some runtime checks. Anyway, I can include your comment here in a generalized form.

rrusak (approved):

wbober (approved):

wbober:
LGTM, I think we will need to iterate over this in following releases, but we can merge.

grochu (approved):
Looks good, small comments I had initially have been resolved already

doc/libraries/nrf_edgeai_obsv.rst
mbarchie: I'll put some numbers and short legend

mbarchie: I'll add formulas, could you review them?

doc/integrations/model_observability.rst
mbarchie: True, I'll put some of your points and remove this untrue statement

mbarchie:
Removing math formulas for observability metrics - @rrusak as their author will prepare formulas as a follow-up task.

Szynkaa (approved):

annwoj:
@wbober @Szynkaa @rrusak @grochu I moved the doc to top-level structure, don't think it fits the integration section. Please let me know in case you have any comments on the new placement/additional changes

doc/model_lifecycle.rst
```diff
@@ -0,0 +1,282 @@
...
+   Today, a model is built into the application binary, so updating it means shipping a full application update.
+   Support for updating only the model, without re-flashing the rest of the application, is planned.
+
+.. rst-class:: numbered-step
+
+Interpreting collected metrics
+==============================
+
+Once the product is in the field, the :ref:`nrf_edgeai_obsv_lib` accumulates metric snapshots from every inference and periodically uploads them off-device for analysis.
+The library ships with a Memfault transport that stages snapshots as a `Memfault Custom Data Recording`_ (CDR) uploaded to nRF Cloud, or you can implement a custom transport over any connectivity solution your product already uses.
```
mbarchie, line 186: @annwoj This was previously a part of separate section no. 4. IMO it could be kept as separate to distinguish collection and analysis. This way we also emphasize that we have a tool for this step - nRF Cloud / Memfault. Analysis can be made somewhere else and it doesn't have to happen continuously.

What's more, removing that point breaks numbering on both diagrams, but this one can be fixed.

doc/model_lifecycle.rst
```diff
@@ -0,0 +1,282 @@
...
+.. _model_observability_hardware_environment:
+
+Revise hardware or the deployment environment
+---------------------------------------------
+
+Some issues that observability data reveals cannot be fixed by changing software alone.
+A failing sensor, an enclosure that distorts the signal, or a deployment environment consistently different from what the product was designed for are hardware or product design problems.
+
+Depending on the root cause, the fix might be replacing or recalibrating the affected sensor, revising the product's mechanical or electrical design, or updating installation guidance for the deployment environment.
+Once you revise your design and introduce improvements, continue observing the product to confirm that the metrics return to expected levels.
```
mbarchie, line 252: @annwoj Could we leave them marked as "a, b, c, ..." or make some other sort of relation between text and diagrams? IMO letters did their job quite well.

For example 1st diagram uses 6a, same for the 2nd diagram and 6a, 6b, 6c, but it doesn't link nowhere in the text.

mbarchie (commented):
@annwoj I'm ok with structural changes and moving it out from Integrations to top level. I put comments regarding points 1-6 in the lifecycle page, I have doubts about linking between diagram and text.

doc/model_lifecycle.rst
```diff
@@ -0,0 +1,282 @@
...
+
+Next steps
+***********
+
+* :ref:`nrf_edgeai_obsv_lib` — Library reference, built-in metrics, Kconfig, and integration steps.
+* :ref:`app_ww_kws` — Reference application with observability enabled.
```
Szynkaa: observavility is configured and enabled when proper build type is selected

doc/model_lifecycle.rst
```diff
@@ -0,0 +1,282 @@
...
+Model observability gives you a structured way to capture these statistics on-device and send them off-device for analysis, so you can tell whether a deployed model is still performing as expected.
+
+The |EAI| implements model observability through the :ref:`nrf_edgeai_obsv_lib`.
+The library works with any inference engine that exposes a class-probability vector, an input-feature vector, or both — depending on which built-in or custom metrics you enable — including the :ref:`nrf_edgeai_lib`, :ref:`Axon NPU <lib_axon>`, and `Edge Impulse`_ deployments.
+
+.. note::
+   Model observability is designed primarily around |EAILib|, which provides the most seamless integration.
+   To get the most out of model observability, consider using |EAILib| for your models.
+
+Root cause categories
```
Szynkaa: I would not mark this as root cause categories, the effects and root causes are now presented like we would have confidence in them. AFAIK we performed some tests, but this analysis is application specific. I prefer the old style where I get feeling that there are paths to search for root cause

doc/model_lifecycle.rst
```diff
@@ -0,0 +1,282 @@
...
+
+While integrating the model, configure the :ref:`nrf_edgeai_obsv_lib` metrics you want to collect in production.
+See :ref:`nrf_edgeai_obsv_lib` for Kconfig options, initialization, and API usage.
+
+.. _model_observability_deployment:
+
+.. rst-class:: numbered-step
+
+Deployment
+========================
```
Szynkaa: match underling to section title length, also lines 184 and 238

doc/model_lifecycle.rst
annwoj: applied

doc/model_lifecycle.rst
annwoj: fixed - at first I didn't notice the correlation between numbers and diagram refs, thanks!

doc/model_lifecycle.rst
annwoj: thanks for additional context! updated to reflect that it's a guideline, not a fixed troubleshooting scenario

annwoj (approved):

mbarchie:
rebase

doc/model_lifecycle.rst
```diff
@@ -0,0 +1,294 @@
...
+Model observability gives you a structured way to capture these statistics on-device and send them off-device for analysis, so you can tell whether a deployed model is still performing as expected.
+
+The |EAI| implements model observability through the :ref:`nrf_edgeai_obsv_lib`.
+The library works with any inference engine that exposes a class-probability vector, an input-feature vector, or both — depending on which built-in or custom metrics you enable — including the :ref:`nrf_edgeai_lib`, :ref:`Axon NPU <lib_axon>`, and `Edge Impulse`_ deployments.
+
+.. note::
+   Model observability is designed primarily around |EAILib|, which provides the most seamless integration.
+   To get the most out of model observability, consider using |EAILib| for your models.
+
+Production investigation paths
```
Szynkaa: page is titled Model lifecycle and observability but first topic after overview is Production investigation paths. I think this need to go below or inside The Edge AI model lifecycle

mbarchie: After a second look - I agree with @Szynkaa.

@annwoj I can think of 3 options:
1) Restore the original "continuous" flow of this section (without subsections) and make it a part of "Overview"
2) Drop "Production investigation paths" and move the rest to "Overview"
3) Drop "Production investigation paths" and move the rest to "5. Analyzing collected data"

annwoj: Let's keep it split into subsections as it is, for improved the readability. However, we don't need the Production investigation paths heading. You can move it to overview or step 5 - wherever you think it fits bettter, I have no strong opinion here

mbarchie: @annwoj I chose option 4 - to move it fairly unchanged (still with subsections) as a separate section, to the end of article, and refer to it from Step 5. Please check if it works for you.

I also added a note about Edge AI Lib not supporting input observabilties yet in both lifecycle and lib pages - as requested by @wbober

annwoj: works for me, thanks!

grochu (approved):

Szynkaa (approved):

PR #145, opened 2026-08-18
Feature/nrf edgeai 3.0.0+new model observabilities

rrusak:
This pull request refactors includes a nrf_edgeai 3.0.0 and enhances the observability configuration and implementation for the wakeword and keyword spotting models in the `ww_kws` application. The changes introduce more granular control over which models have observability enabled, centralize observability handling, and update documentation and configuration files accordingly. Additionally, several generated model files are updated to support new feature scaling interfaces.

Key changes include:

### New nrf_edgeai version
Update nrf_edgeai to 3.0.0 version

### Observability configuration and implementation

* Split the monolithic `CONFIG_MODELS_OBSERVABILITY` Kconfig option into two separate options: `CONFIG_MODELS_OBSERVABILITY_WW` (for the wakeword model) and `CONFIG_MODELS_OBSERVABILITY_KWS` (for the keyword spotting model), allowing independent enabling of observability for each model. An umbrella `CONFIG_MODELS_OBSERVABILITY` is now auto-selected if either is enabled.
* Updated `observability.conf` to enable only the wakeword observability by default and added detailed comments about configuration and memory usage. All built-in metrics are now enabled for both models, with configurable histogram bin numbers and other metric parameters.
* Refactored `src/kws/kws.c` to use the new `model_obsv` shared module for observability, removing direct metric and Memfault context management from the model file. Observability is now conditionally compiled only if `CONFIG_MODELS_OBSERVABILITY_KWS` is enabled. [[1]](diffhunk://#diff-3571ce1d2ba90077d9de9e0114e5049334a0a50a534725eab000d95b54c32865R12-R19) [[2]](diffhunk://#diff-3571ce1d2ba90077d9de9e0114e5049334a0a50a534725eab000d95b54c32865L52-R66) [[3]](diffhunk://#diff-3571ce1d2ba90077d9de9e0114e5049334a0a50a534725eab000d95b54c32865L74-R81) [[4]](diffhunk://#diff-3571ce1d2ba90077d9de9e0114e5049334a0a50a534725eab000d95b54c32865L119-R99)
* Updated CMake configuration to include the `model_obsv.c` source file only when observability is enabled.

### Documentation updates

* Revised the `README.rst` to reflect the new observability configuration, clarifying how to enable observability for each model and describing the centralized role of the `model_obsv.c` module. [[1]](diffhunk://#diff-0d8dac1192e38b0c42fd1722bc666555013d8e86b43afb6e020744abbb30d4beL44-R48) [[2]](diffhunk://#diff-0d8dac1192e38b0c42fd1722bc666555013d8e86b43afb6e020744abbb30d4beL184-R184)

### Model interface updates

* Updated generated model files for all supported platforms to add a new `scale_features` interface . constant. [[1]](diffhunk://#diff-fce95c0f1b35d8e0f7e0d0fae9835353ed0f8d788c503f1dee8ea6d82ce3890dL10-R10) [[2]](diffhunk://#diff-fce95c0f1b35d8e0f7e0d0fae9835353ed0f8d788c503f1dee8ea6d82ce3890dL340-R341) [[3]](diffhunk://#diff-fce95c0f1b35d8e0f7e0d0fae9835353ed0f8d788c503f1dee8ea6d82ce3890dR388) [[4]](diffhunk://#diff-3bb89a09ff303709423cdb10995a41dc6277e4f14b3852c7ab3ab24ea6607befL15-R15) [[5]](diffhunk://#diff-3bb89a09ff303709423cdb10995a41dc6277e4f14b3852c7ab3ab24ea6607befL247-R248) [[6]](diffhunk://#diff-3bb89a09ff303709423cdb10995a41dc6277e4f14b3852c7ab3ab24ea6607befR294) [[7]](diffhunk://#diff-e0eb2406d88a7ad9704fef68b63b9a68f32cd42b8e4fe5e670f33700e5f48425L10-R10) [[8]](diffhunk://#diff-e0eb2406d88a7ad9704fef68b63b9a68f32cd42b8e4fe5e670f33700e5f48425L361-R362) [[9]](diffhunk://#diff-e0eb2406d88a7ad9704fef68b63b9a68f32cd42b8e4fe5e670f33700e5f48425R409) [[10]](diffhunk://#diff-1a1268bd1df7084e2e9f30e1771d23605b6cdb7cbaf5a1db2b8f293254a4ec98L10-R10) [[11]](diffhunk://#diff-1a1268bd1df7084e2e9f30e1771d23605b6cdb7cbaf5a1db2b8f293254a4ec98L290-R291) [[12]](diffhunk://#diff-1a1268bd1df7084e2e9f30e1771d23605b6cdb7cbaf5a1db2b8f293254a4ec98R338)


include/nrf_edgeai/rt/nrf_edgeai_types.h
```diff
@@ -130,10 +142,40 @@ typedef union nrf_edgeai_rt_version_u
  */
 typedef struct nrf_edgeai_metadata_s
 {
-    const char*             p_solution_id; /**< Solution ID string */
-    nrf_edgeai_rt_version_t version;       /**< Solution runtime version */
+    /**< Solution ID string */
```
Szynkaa: don't use inline docs marking for non-inline docs
same below

applications/ww_kws/src/obsv/model_obsv.h
```diff
@@ -0,0 +1,127 @@
...
+
+#define _MODEL_OBSV_WORDS(bytes) ((bytes) / sizeof(uint32_t))
+
+/**
+ * @brief One model's observability state: context + every enabled metric's
+ *        descriptor and (uint32_t-aligned) counter storage.
+ *
+ * Declare one static instance per model and pass it to @ref model_obsv_init.
+ */
+struct model_obsv {
```
Szynkaa: this increase memory usage for not used storage in wakeword and create need for synthesized [1 - p, p] vector, I see it was done for code deduplication, but I think it hides how to configure observability in clients application

applications/ww_kws/Kconfig
```diff
@@ -64,27 +64,41 @@ config KWS_EMA_ALPHA
...
+	depends on NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS >= 2 || !MODELS_OBSERVABILITY_WW
 	help
-	  Wire observability into the bundled models. Each model file owns its observability context
-	  and Memfault transport binding.
+	  Collect observability metrics for the keyword spotting model. Enabling it together with
+	  MODELS_OBSERVABILITY_WW (wakeword-gated mode runs both models) registers two Memfault
+	  contexts, so NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS must be >= 2.
 
-if MODELS_OBSERVABILITY
+config MODELS_OBSERVABILITY
```
Szynkaa, line 87: this symbol can be removed, as it is only used as MODELS_OBSERVABILITY_MDS dependency

applications/gesture_recognition/src/nrf_edgeai_generated/nrf54l15tag/nrf_edgeai_user_model.c
```diff
@@ -384,6 +385,7 @@ static nrf_edgeai_t nrf_edgeai_ = {
 	.interfaces.input_init = NN_INPUT_INIT_INTERFACE,
 	.interfaces.feed_inputs = NN_INPUT_FEED_INTERFACE,
 	.interfaces.process_features = NN_PROCESS_FEATURES_INTERFACE,
+    .interfaces.scale_features = NN_SCALE_FEATURES_INTERFACE,
```
mbarchie: whitespace formatting is updated model files is inconsistent (tabs vs spaces)

applications/ww_kws/src/kws/kws.c
```diff
@@ -49,17 +50,20 @@ static const struct keyword_detection_ctx keyword_detection_ctxs[] = {
...
 
-#if IS_ENABLED(CONFIG_MODELS_OBSERVABILITY)
+#if IS_ENABLED(CONFIG_MODELS_OBSERVABILITY_KWS)
 
-static nrf_edgeai_obsv_ctx_t kws_ctx;
+/* Mel feature vector length produced by the model DSP front end. Sizes the
+ * FEATURES-stream metric storage; validated at runtime against
+ * nrf_edgeai_dsp_features_ctx()->overall_num.
+ */
+#define KWS_NUM_FEATURES 40
```
mbarchie: Is the comment correct? I couldn't find any runtime check against `overall_num`. Same for WW.

applications/ww_kws/src/kws/kws.c
```diff
@@ -196,6 +174,27 @@ int kws_process(uint8_t *const audio_buffer, const uint16_t num_samples,
...
 
+#if IS_ENABLED(CONFIG_MODELS_OBSERVABILITY_KWS)
+	/* Extract the mel feature vector before inference and feed it to the
+	 * FEATURES-stream metrics (mel energy / spectral descriptors). run_inference
+	 * reuses these features, so the explicit call adds no extra DSP work.
+	 */
+	err = nrf_edgeai_process_features(kws_model);
+	if (err == NRF_EDGEAI_ERR_INPROGRESS) {
+		/* Feature window not complete yet. */
+		return -EBUSY;
```
mbarchie, line 157: Is `NRF_EDGEAI_ERR_INPROGRESS` expected to be returned when window is not full? Or `NRF_EDGEAI_ERR_WRONG_STATE` - as I understand it from description of `nrf_edgeai_process_features()`.

I have the same doubt about `nrf_edgeai_run_inference()` looking at its description.

applications/ww_kws/observability.conf
```diff
@@ -4,20 +4,49 @@
...
 
-# Auto collect metrics
+# Auto collect metrics. The collect encodes all contexts into an
+# NRF_EDGEAI_OBSV_ENCODE_LIST_BUFSZ buffer on the system workqueue stack; that
+# buffer grows with metrics x bins x classes x MAX_CONTEXTS (~8 KB for the full
+# set at 16 bins, 12 classes, 2 contexts), so the workqueue stack is raised well
+# above the default. If RAM is tight, lower the *_BIN_NUM values above.
 CONFIG_NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT=y
-CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=4096
+CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=12288
```
mbarchie, line 52: With only WW observability enabled in this .conf file, there's no need to set `CONFIG_NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS` to 2 and increase `CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE` that much.

I get it's to make it easier to play with enabling/disabling WW/KWS observability, without needing to tweak other options. However, in this case, IMO it would be better to deliver this sample with observability enabled for both models by default.

applications/ww_kws/README.rst
```diff
@@ -181,7 +181,7 @@ Configuration files
 ===================
 
 The application provides predefined :file:`observability.conf` configuration file for enabling observability.
-This file enables the ``CONFIG_MODELS_OBSERVABILITY`` Kconfig option, Bluetooth LE, Memfault Diagnostic Service and required dependencies to provide observability for bundled models.
+This file enables observability for both bundled models (``CONFIG_MODELS_OBSERVABILITY_WW`` and ``CONFIG_MODELS_OBSERVABILITY_KWS``), Bluetooth LE, Memfault Diagnostic Service and required dependencies. Enable only one of the two options to observe just the wakeword or just the keyword spotting model.
```
mbarchie: This doesn't match what's in `observability.conf` - only WW is enabled. Plus new sentence should go into new line.
```suggestion
This file enables observability for both bundled models (``CONFIG_MODELS_OBSERVABILITY_WW`` and ``CONFIG_MODELS_OBSERVABILITY_KWS``), Bluetooth LE, Memfault Diagnostic Service and required dependencies.
Enable only one of the two options to observe just the wakeword or just the keyword spotting model.
```

applications/ww_kws/Kconfig
```diff
@@ -64,27 +64,41 @@ config KWS_EMA_ALPHA
...
+
+config MODELS_OBSERVABILITY_KWS
+	bool "Edge AI observability for the keyword spotting model"
 	depends on NRF_EDGEAI_OBSV
 	depends on NRF_EDGEAI_OBSV_MEMFAULT
-	depends on NRF_EDGEAI_OBSV_METRIC_PROBS_DISTRIBUTION
-	depends on NRF_EDGEAI_OBSV_METRIC_TRANSITION_MATRIX || APP_MODE_WW_ONLY
-	depends on NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS >= 2 || !APP_MODE_WW_GATED_KWS
+	depends on APP_MODE_WW_GATED_KWS || APP_MODE_KWS_ONLY
+	depends on NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS >= 2 || !MODELS_OBSERVABILITY_WW
```
mbarchie, line 81: Doesn't `MODELS_OBSERVABILITY_WW` need similar dependency on number of contexts?

applications/ww_kws/src/obsv/model_obsv.h
mbarchie: I agree with @Szynkaa comment - this code deduplication comes with big cost in memory and I guess it's not how users would exactly use it in their apps. It's better for us, who are testing observability feature.

I think metrics should be enabled per-model and sized appropriately. One alternative idea would be some config array, so that `model_obsv_init()` could iterate over it and register them for specific models? I think buffers could be also statically allocated based on some config, but it would require use of some X-macro probably - which would increase complexity and reduce readability of sample application.

applications/ww_kws/src/kws/kws.c
rrusak: For ww/kws we have internal 3x40 discrete mel features window, each nrf_edgeai_process_features computes 1x40 mel vector, so we will return NRF_EDGEAI_ERR_INPROGRESS if 3x40 has not yet been collected

applications/ww_kws/src/kws/kws.c
rrusak: Yeah it is inside nrf_edgeai runtime library .... :) I'll remove this comment

applications/ww_kws/Kconfig
rrusak: The bare MODELS_OBSERVABILITY symbol has two functional uses:
- target_sources_ifdef(CONFIG_MODELS_OBSERVABILITY app PRIVATE src/obsv/model_obsv.c) - this is what compiles the shared model_obsv.c module. This is actually its primary job from my point of view
- Kconfig — depends on MODELS_OBSERVABILITY (the _MDS dependency).

Why I think that MODELS_OBSERVABILITY_MDS  != MODELS_OBSERVABILITY and it's a strict subset than equivalent to "observability is enabled". MDS is y only when observability is on AND the full BLE-MDS drain stack is present. It's much narrower than "observability is enabled." If in  observability.conf MDS happens to be y (BT + auto-collect are on), so it would seem to work - but it silently couples the shared module's existence to the BLE stack, and breaks the moment someone uses obsv **without MDS**. 

I think the module must be present exactly when its symbols are referenced = when obsv is on for any model = MODELS_OBSERVABILITY (= WW || KWS). That's what the umbrella is for, and it's why the split is:

model_obsv.c - gated on MODELS_OBSERVABILITY (obsv used at all).
ble_mds.c - gated on MODELS_OBSERVABILITY_MDS (the BLE-MDS drain path specifically).
Two different conditions, two different files. ble_mds.c is the only thing that legitimately keys off _MDS.

Szynkaa: I agree that MODELS_OBSERVABILITY have different semantics than MODELS_OBSERVABILITY_MDS. I missed the cmake dependency on MODELS_OBSERVABILITY
My comment was invalid

applications/ww_kws/observability.conf
rrusak: or make it conditional?

```
config SYSTEM_WORKQUEUE_STACK_SIZE
	default 12288 if MODELS_OBSERVABILITY_WW && MODELS_OBSERVABILITY_KWS 
	default 8192  if MODELS_OBSERVABILITY_KWS                            
	default 4096  if MODELS_OBSERVABILITY_WW  
```

applications/ww_kws/src/obsv/model_obsv.h
rrusak: Got it. Do we need to change this in *this* specific release? Figure out a "clean" structure and an initialization method for it isn't something that I can do quickly.

applications/ww_kws/README.rst
```diff
@@ -41,11 +41,11 @@ Observability
 =============
 
 The application collects runtime observability metrics using the :ref:`nrf_edgeai_obsv_lib`.
-Enable the ``CONFIG_MODELS_OBSERVABILITY`` Kconfig option to wire metrics into the bundled models.
-Each model stage owns its observability context and Memfault transport binding in the :file:`src/ww/wakeword.c` and :file:`src/kws/kws.c` files.
+Enable ``CONFIG_MODELS_OBSERVABILITY_WW``, ``CONFIG_MODELS_OBSERVABILITY_KWS``, or both to wire metrics into the wakeword model, the keyword spotting model, or both.
+The shared :file:`src/obsv/model_obsv.c` module owns each model's observability context, registers every enabled built-in metric, and binds the Memfault transport; the model files (:file:`src/ww/wakeword.c` and :file:`src/kws/kws.c`) only feed samples to it.
```
annwoj:
```suggestion
Enable metrics for each model independently (either option, or both):

* ``CONFIG_MODELS_OBSERVABILITY_WW`` - Wakeword model
* ``CONFIG_MODELS_OBSERVABILITY_KWS`` - Keyword spotting model

The :file:`src/obsv/model_obsv.c` module does all the observability work.
For each enabled model, it owns the observability context, registers the built-in metrics selected in :file:`observability.conf`, and binds the Memfault transport.
The model files (:file:`src/ww/wakeword.c` and :file:`src/kws/kws.c`) only feed samples to this module.

Because both models use the same metric set, the single output of the wakeword model is expanded into a synthetic two-class ``[1 - p, p]`` distribution before the probability metrics are updated.
```

applications/ww_kws/README.rst
```diff
@@ -41,11 +41,11 @@ Observability
...
 
-The wakeword stage only registers the :ref:`nrf_edgeai_obsv_metrics_built_in_probability` metric.
-The keyword spotting stage registers the :ref:`nrf_edgeai_obsv_metrics_built_in_probability` and :ref:`nrf_edgeai_obsv_metrics_built_in_transition` metrics.
+Both models register the same metric set: whichever built-in metrics are enabled at build time (see :file:`observability.conf`).
+The wakeword model has a single output, so it is expanded into a synthetic two-class ``[1 - p, p]`` distribution before the probability metrics, which lets the same set apply to it as to the multi-class keyword spotting model.
```
annwoj:
```suggestion
```

doc/libraries/nrf_edgeai_changelog.rst
```diff
@@ -9,6 +9,50 @@ nRF Edge AI Library changelog
...
+Known issues
+============
+
+There are no critical known issues identified for this release.
+
+Compatibility
+=============
+
+* Nordic Edge AI Lab solutions version: 3.0.0
+* Axon driver version: 1.3.0 - 1.5.0
```
annwoj: Maybe instead of "breaking changes" we should change it into "migration notes"? something we usually do in ncs.

```suggestion
Release v3.0.0 (17 August 2026)
*******************************

This release is tagged as ``NRF-EDGEAI-RELEASE-3.0.0`` (internal release commit ``6dfb365bca9400f463b8c16773887b6d6fa39550``).

This is a major release.
The runtime major version was raised from 2 to 3, which makes it incompatible with solutions generated for the v2.x runtime.

* This release contains the following breaking changes:

  * Solutions and applications built for the 2.x runtime do not work with this release.
    Retrain your model and export a new solution with Nordic Edge AI Lab 3.0.0.
    :c:func:`nrf_edgeai_is_runtime_compatible` compares the major version of the runtime library against the solution, so a solution created for a 2.x runtime causes :c:func:`nrf_edgeai_init` to fail with ``NRF_EDGEAI_ERR_INCOMPATIBLE``.
  * Feature processing is now split into two separate pipeline stages: DSP feature extraction and feature scaling.
    The ``nrf_edgeai_interfaces_t`` runtime interface structure gained a new mandatory ``scale_features`` member of type ``nrf_edgeai_iface_scale_features_t``, which every solution context must provide.
    Solutions exported with Nordic Edge AI Lab 3.0.0 populate this member automatically.
  * The combined ``nrf_edgeai_process_features_<mode>_<input>_<output>`` interface family was removed and replaced by the following:
    * ``process_features``, which now provides DSP extraction only, through ``nrf_edgeai_process_features_dsp_i8()``, ``nrf_edgeai_process_features_dsp_i16()``, ``nrf_edgeai_process_features_dsp_f32()``, and ``nrf_edgeai_process_features_empty()``.
    * The new ``nrf_edgeai_scale_features_*`` family declared in :file:`nrf_edgeai_scale_features.h`, which handles scaling.
    * Renamed scaling mode prefixes, where ``scale_vector_*`` is now ``input_vector_*``, and ``scale_window_*`` is now ``input_window_*``.
  * The ``nrf_edgeai_t`` runtime context gained a ``state`` member, which changes the structure layout.
    Applications and exported solutions must be recompiled against the 3.0.0 headers.
   
* Added:

  * The public :c:func:`nrf_edgeai_process_features` API, allowing DSP feature extraction to be triggered independently of model inference.
  * Runtime state tracking through ``nrf_edgeai_state_t``, with composite readiness masks and the following state bit definitions:

    * ``NRF_EDGEAI_STATE_RT_INITIALIZED``
    * ``NRF_EDGEAI_STATE_RT_INPUTS_COLLECTED``
    * ``NRF_EDGEAI_STATE_RT_FEATURES_PROCESSED``
    * ``NRF_EDGEAI_STATE_RT_INFERENCE_COMPLETED``

    The runtime now validates its state on entry to the input, feature processing, and inference stages.

  * Two error codes for state violations: ``NRF_EDGEAI_ERR_UNINITIALIZED`` (-8) and ``NRF_EDGEAI_ERR_WRONG_STATE`` (-9).
  * :c:func:`nrf_edgeai_dsp_features_ctx` to obtain read-only access to the DSP feature extraction context, so computed features can be used from the application.

* Updated:

  * Feature processing was split into two separate pipeline stages: DSP feature extraction and feature scaling.
  * :c:func:`nrf_edgeai_run_inference` keeps the previous single-call workflow working.
    If features have not been processed yet, it invokes :c:func:`nrf_edgeai_process_features` internally and propagates its status code unchanged.

* Fixed:

  * :c:func:`nrf_edgeai_feed_inputs` now rejects calls on a context that was never successfully initialized, returning ``NRF_EDGEAI_ERR_UNINITIALIZED`` instead of forwarding the data to an uninitialized input window context.

Known issues
============

There are no critical known issues identified for this release.

Compatibility
=============

* Nordic Edge AI Lab solutions version: 3.0.0
* Axon driver version: 1.3.0 to 1.5.0
```

annwoj (changes requested):

applications/ww_kws/Kconfig
rrusak: The KWS guard MAX_CONTEXTS >= 2 || !MODELS_OBSERVABILITY_WW is that condition written from KWS's side. Because it's a dependency of KWS, it's evaluated whenever KWS would be y:

If KWS=y, then (>=2 || !WW) must hold - NOT(WW && KWS && <2).
If KWS=n, the illegal combo can't exist (it needs KWS).

So the state WW && KWS && MAX_CONTEXTS<2 is unreachable with the guard on KWS alone. Why we need to check NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS  in WW?

Maybe better to add to Kconfig something like:
```
config NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS
	default 2 if MODELS_OBSERVABILITY_WW && MODELS_OBSERVABILITY_KWS
	default 1 if MODELS_OBSERVABILITY
```

mbarchie: I agree the current one guards against misconfiguration from logical point of view and the other check would be redundant. But it doesn't seem intuitive from UX point of view - misconfiguration of contexts always pops up as KWS failure.

I think your Kconfig fix is fine.

applications/ww_kws/observability.conf
mbarchie: It would work, but I'm not convinced that the same approach as for `CONFIG_NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS` is the best one.

With your proposal `SYSTEM_WORKQUEUE_STACK_SIZE` would be hidden in Kconfig and pretending to depend just from observability. However, it's a shared, global resource, so the correct size is the max of everyone's needs, not just the observability. If we compute it purely from `MODELS_OBSERVABILITY_WW/_KWS`, that's an implicit claim that these are the only two things that matter for this stack. A developer re-using this sample naturally reads `observability.conf` and tunes what's there and an explicit `CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE` keeps that setting visible for them to bump

doc/libraries/nrf_edgeai_changelog.rst
rrusak: Currently, there is no "regenerate solution" option in Nordic Edge AI Lab; the only way is for the user to retrain everything from scratch. 
Is it even necessary to write this migration if the add-on is still in the experimental phase? I wouldn't want the user to manually edit the model's internal context.

annwoj: Got it - thought we actually do have migration steps and this is the desired solution. 
But in this case it's as you mention, we don't need the migration steps.
I updated the edit - please check

applications/ww_kws/Kconfig
```diff
@@ -99,6 +99,18 @@ config MODELS_OBSERVABILITY_MDS
 	depends on BT_SMP
 	depends on NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT
 
+# Observability-owned memory sizing — these are private to observability, so they
+# scale automatically with the enabled models. Placed before "source Kconfig.zephyr"
+# so these app defaults take precedence over the library defaults. Override in a
+# .conf for a custom model whose keyword count differs from the bundled 12-class one.
+config NRF_EDGEAI_OBSV_MAX_CLASSES
+	default 12 if MODELS_OBSERVABILITY_KWS
+	default 2 if MODELS_OBSERVABILITY_WW
```
Szynkaa: I think this should stay in .conf. Value depends not only on which models have observability enabled (y in .conf already), but also on the model itself

applications/ww_kws/observability.conf
```diff
@@ -38,14 +41,15 @@ CONFIG_NRF_EDGEAI_OBSV_METRIC_MEL_SPECTRAL_DESC=y
...
-# above the default. If RAM is tight, lower the *_BIN_NUM values above.
 CONFIG_NRF_EDGEAI_OBSV_MEMFAULT_AUTO_COLLECT=y
+
+# System workqueue stack — a SHARED, global stack (Memfault packetizer, logging,
+# BLE, ...), so it is set explicitly here rather than derived; its correct size is
+# the max over all its users. This value covers the observability encode buffer for
+# the default scope (both models = 2 contexts x 12 classes) plus headroom. Lower it
+# to ~8192 for the keyword model only or ~4096 for the wakeword model only; raise it
+# if you add other workqueue-heavy work. The observability Memfault library
+# BUILD_ASSERTs its own floor, so an under-sized value fails the build.
```
Szynkaa: brieffer, no need to comment about deriving size from other kconfigs

applications/ww_kws/src/obsv/model_obsv.h
Szynkaa: not addressed

Szynkaa (commented):
the commits needs squashing
- lib update with models alignment
- (optional) separate lib changelog update
- observability update in KWS

doc/libraries/nrf_edgeai_changelog.rst
```diff
@@ -9,6 +9,60 @@ nRF Edge AI Library changelog
...
+Release v3.0.0 (17 August 2026)
+*******************************
+
+This release is tagged as ``NRF-EDGEAI-RELEASE-3.0.0`` (internal release commit ``6dfb365bca9400f463b8c16773887b6d6fa39550``).
+This is a major release.
+
+The runtime major version was raised from 2 to 3, which makes it incompatible with solutions generated for the v2.x runtime.
+* This release contains the following breaking changes:
+  * Solutions and applications built for the 2.x runtime do not work with this release.
+    Retrain your model and export a new solution with Nordic Edge AI Lab 3.0.0.
```
annwoj, line 23:
```suggestion

* This release contains the following breaking changes:

  * Solutions and applications built for the 2.x runtime do not work with this release.
    Retrain your model and export a new solution with Nordic Edge AI Lab 3.0.0.
```

doc/libraries/nrf_edgeai_changelog.rst
```diff
@@ -9,6 +9,60 @@ nRF Edge AI Library changelog
...
+    :c:func:`nrf_edgeai_is_runtime_compatible` compares the major version of the runtime library against the solution, so a solution created for a 2.x runtime causes :c:func:`nrf_edgeai_init` to fail with ``NRF_EDGEAI_ERR_INCOMPATIBLE``.
```
annwoj, line 24: I think the indent here should be aligned with what's above(?) Won't render correctly otherwise

doc/libraries/nrf_edgeai_changelog.rst
```diff
@@ -9,6 +9,60 @@ nRF Edge AI Library changelog
...
+    * ``NRF_EDGEAI_STATE_RT_INITIALIZED``
+    * ``NRF_EDGEAI_STATE_RT_INPUTS_COLLECTED``
+    * ``NRF_EDGEAI_STATE_RT_FEATURES_PROCESSED``
+    * ``NRF_EDGEAI_STATE_RT_INFERENCE_COMPLETED``
+
+    The runtime now validates its state on entry to the input, feature processing, and inference stages.
+  * Two error codes for state violations: ``NRF_EDGEAI_ERR_UNINITIALIZED`` (-8) and ``NRF_EDGEAI_ERR_WRONG_STATE`` (-9).
+  * :c:func:`nrf_edgeai_dsp_features_ctx` to obtain read-only access to the DSP feature extraction context, so computed features can be used from the application.
+
+* Updated:
```
annwoj, line 49:
```suggestion
* Updated:

```

doc/libraries/nrf_edgeai_changelog.rst
```diff
@@ -9,6 +9,60 @@ nRF Edge AI Library changelog
...
+  * Feature processing was split into two separate pipeline stages: DSP feature extraction and feature scaling.
+  * :c:func:`nrf_edgeai_run_inference` keeps the previous single-call workflow working.
+    If features have not been processed yet, it invokes :c:func:`nrf_edgeai_process_features` internally and propagates its status code unchanged.
+
+* Fixed:
```
annwoj, line 55:
```suggestion
* Fixed:

```

applications/ww_kws/src/obsv/model_obsv.h
rrusak: Work in progress

applications/ww_kws/src/ww/wakeword.c
```diff
@@ -134,6 +94,27 @@ int ww_process(uint8_t *const audio_buffer, const uint16_t num_samples, bool *co
 		return -EPERM;
 	}
 
+#if IS_ENABLED(CONFIG_MODELS_OBSERVABILITY_WW)
+	/* Extract the mel feature vector before inference for the FEATURES-stream
+	 * metrics. run_inference reuses these features, so the explicit call adds no
+	 * extra DSP work.
+	 */
+	err = nrf_edgeai_process_features(ww_model);
+	if (err == NRF_EDGEAI_ERR_INPROGRESS) {
```
grochu, line 103: Shouldn't the `NRF_EDGEAI_ERR_WRONG_STATE` code be checked for a case that feature window is not complete yet? API doc: https://github.com/nrfconnect/sdk-edge-ai/pull/145/changes#diff-601d8fd4bb578bb6f2198666ccbc9bf852a2010da7e785f5acdbc8ef4cb014f2R75 - or maybe the header requires alignment?

applications/ww_kws/observability.conf
```diff
@@ -4,20 +4,54 @@
...
+# Observability memory sizing These values cover the default scope
+# (both models): MAX_CLASSES is the largest observed class count (keyword model = 12,
+# the wakeword model needs only its synthesized 2), MAX_CONTEXTS is one Memfault
+# context per observed model. Use 2 / 1 for the wakeword model only, 12 / 1 for the
+# keyword model only, and change MAX_CLASSES for a custom model.
+CONFIG_NRF_EDGEAI_OBSV_MAX_CLASSES=12
+CONFIG_NRF_EDGEAI_OBSV_MEMFAULT_MAX_CONTEXTS=2
+
+# All built-in metrics, applied to both models (the wakeword model's single output
+# is expanded to a synthetic 2-class distribution in wakeword.c). Histograms at 16
```
grochu: Now it's done in `ww_obsv.c` not in `wakeword.c` - maybe remove the source file name from the comment?

applications/ww_kws/src/kws/kws_obsv.c
```diff
@@ -0,0 +1,170 @@
...
+	if (err) {
+		LOG_ERR("Memfault transport init failed (err %d)", err);
+	}
+
+	return err;
+}
+
+void kws_obsv_update_features(const float *feats, uint16_t n)
+{
+	int err = nrf_edgeai_obsv_update_features(&ctx, feats, n);
```
grochu, line 161: Shouldn't `n` be checked against `KWS_OBSV_FEATURES`?

applications/ww_kws/src/ww/ww_obsv.c
```diff
@@ -0,0 +1,179 @@
...
+	if (err) {
+		LOG_ERR("Memfault transport init failed (err %d)", err);
+	}
+
+	return err;
+}
+
+void ww_obsv_update_features(const float *feats, uint16_t n)
+{
+	int err = nrf_edgeai_obsv_update_features(&ctx, feats, n);
```
grochu, line 168: As https://github.com/nrfconnect/sdk-edge-ai/pull/145/changes#r3958873402, shouldn't `n` be checked against `WW_OBSV_FEATURES`?

applications/ww_kws/src/ww/wakeword.c
rrusak: but this statement will show if window was incomplete when calling this function
else if (err) {
		LOG_ERR("Failed to process features (err %d)", err);
		return -EPERM;
	}

grochu: That's right, maybe just the comment below is misleading.

grochu (approved):

applications/ww_kws/src/ww/ww_obsv.h
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+#ifndef WW_OBSV_H_
+#define WW_OBSV_H_
+
+#include <stdint.h>
+
+#include <nrf_edgeai/nrf_edgeai.h>
+
+/** @brief Create the observability context and metrics for the wakeword model. */
```
mbarchie, line 14: No param/return in doxygen for functions in `ww_obsv.h` and `kws_obsv.h`

mbarchie (commented):
a few compliance checks failing + history of commits need to be cleaned up before merging

PR #153, opened 2026-09-15
doc: add known issues filtering

annwoj:
Added known issues filtering for better UX


Szynkaa:
Similar feature was already in nrf: `page_filter` extension with `version-filter` directive

